PR #4, opened 2022-04-06
Update `tests/run.py` for Jenkins server

DavdGao:
- Import current path in `tests/run.py`
- Report failures by `exit(1)`


joneswong (approved):
We have to merge this changes now. We shall avoid appending the path of this package.

PR #9, opened 2022-04-11
fix gan cra loss_batch-> loss_task bug

Osier-Yi:
fix the class representative attack bug: loss_batch -> loss_task


xieyxclack (approved):
approve

PR #11, opened 2022-04-11
[Feature] Add dropout and log training metric.

rayrayraykk:
Add dropout option for CNN and NLP model; 
Add training metrics to logs;
Add instructions for the application version of FederatedScope.

### DIFF
--- a/federatedscope/core/worker/client.py
+++ b/federatedscope/core/worker/client.py
@@ -254,21 +254,26 @@ class Client(Worker):
         self.state = message.state
         if message.content != None:
             self.trainer.update(message.content)
-        test_metrics = self.trainer.evaluate()
-        for key in test_metrics:
-            logging.info(
-                'Client #{:d}: (Evaluation (test set) at Round #{:d}) {:s} is {:.6f}'
-                .format(self.ID, self.state, key, test_metrics[key]))
-        val_metrics = self.trainer.validate()
-        for key in val_metrics:
-            logging.info(
-                'Client #{:d}: (Evaluation (val set) at Round #{:d}) {:s} is {:.6f}'
-                .format(self.ID, self.state, key, val_metrics[key]))
-        train_metrics = self.trainer.evaluate(target_data_split_name='train')
-        for key in train_metrics:
-            logging.info(
-                'Client #{:d}: (Evaluation (train set) at Round #{:d}) {:s} is {:.6f}'
-                .format(self.ID, self.state, key, train_metrics[key]))
+        test_metrics, val_metrics, train_metrics = {}, {}, {}
+        if 'test' in self._cfg.eval.split:
+            test_metrics = self.trainer.evaluate()
+            for key in test_metrics:
+                logging.info(
+                    'Client #{:d}: (Evaluation (test set) at Round #{:d}) {:s} is {:.6f}'
+                    .format(self.ID, self.state, key, test_metrics[key]))
+        if 'val' in self._cfg.eval.split:
+            val_metrics = self.trainer.validate()
+            for key in val_metrics:
+                logging.info(
+                    'Client #{:d}: (Evaluation (val set) at Round #{:d}) {:s} is {:.6f}'
+                    .format(self.ID, self.state, key, val_metrics[key]))
+        if 'train' in self._cfg.eval.split:
+            train_metrics = self.trainer.evaluate(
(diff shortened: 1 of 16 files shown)


federatedscope/core/worker/client.py
```diff
@@ -264,7 +264,12 @@ def callback_funcs_for_evaluate(self, message: Message):
             logging.info(
                 'Client #{:d}: (Evaluation (val set) at Round #{:d}) {:s} is {:.6f}'
                 .format(self.ID, self.state, key, val_metrics[key]))
-        metrics = {**test_metrics, **val_metrics}
+        train_metrics = self.trainer.evaluate(target_data_split_name='train')
```
joneswong, line 267: Is it possible to allow users to determine whether to conduct evaluation on a specific split?

joneswong, line 267: Is it possible to allow users to determine whether to conduct evaluation on a specific split?

joneswong (commented):
good job. It is helpful to have dropout layers. please hv a look at the other comments.

federatedscope/core/worker/client.py
rayrayraykk: I add `cfg.eval.split` for users to determine whether to conduct evaluation on a specific split

federatedscope/core/worker/client.py
```diff
@@ -254,17 +254,27 @@ def callback_funcs_for_evaluate(self, message: Message):
...
-        metrics = {**test_metrics, **val_metrics}
+        test_metrics, val_metrics, train_metrics = {}, {}, {}
+        if 'test' in self._cfg.eval.split:
+            test_metrics = self.trainer.evaluate()
+            for key in test_metrics:
+                logging.info(
+                    'Client #{:d}: (Evaluation (test set) at Round #{:d}) {:s} is {:.6f}'
+                    .format(self.ID, self.state, key, test_metrics[key]))
+        if 'val' in self._cfg.eval.split:
+            val_metrics = self.trainer.validate()
```
joneswong, line 265: @yxdyc @DavdGao  Is is possible to unify the interface, that is to say, using `trainer.evaluate(target_data_split_name='valid')` in place of this

federatedscope/core/worker/client.py
```diff
@@ -254,17 +254,27 @@ def callback_funcs_for_evaluate(self, message: Message):
...
-                'Client #{:d}: (Evaluation (test set) at Round #{:d}) {:s} is {:.6f}'
-                .format(self.ID, self.state, key, test_metrics[key]))
-        val_metrics = self.trainer.validate()
-        for key in val_metrics:
-            logging.info(
-                'Client #{:d}: (Evaluation (val set) at Round #{:d}) {:s} is {:.6f}'
-                .format(self.ID, self.state, key, val_metrics[key]))
-        metrics = {**test_metrics, **val_metrics}
+        test_metrics, val_metrics, train_metrics = {}, {}, {}
+        if 'test' in self._cfg.eval.split:
```
joneswong, line 258: If @DavdGao @yxdyc  can unify the interfaces of trainer, we'd better use a for loop for these splits.

federatedscope/core/worker/client.py
DavdGao: @joneswong , it is feasible to add a new option `valid` for the parameter `target_data_split_name`, and I will push it in a new pr

yxdyc: We have reserved such interfaces (see [evaluate](https://github.com/alibaba/FederatedScope/blob/master/federatedscope/core/trainers/trainer.py#L170) and [valid](https://github.com/alibaba/FederatedScope/blob/master/federatedscope/core/trainers/trainer.py#L185)).

rayrayraykk: > 

So can we replace the function `validate` with `evaluate(target_data_split_name='val')`?

yxdyc: > > 
> 
> So can we replace the function `validate` with `evaluate(target_data_split_name='val')`?

After further discussion, we agree that there is no need to reserve both `evaluate()` and `validate()` interfaces, since the behaviors when doing validation should be the same as the one for evaluation. 
Please @rayrayraykk unify them in this PR by the way.

joneswong (approved):
approve

Issue #12, opened 2022-04-12
`federate.sample_client_rate=xx` will result in no output when conducting FedAvg on Cora.


rayrayraykk:
How to reproduce:
`python federatedscope/main.py --cfg federatedscope/gfl/baseline/fedavg_gnn_node_fullbatch_citation.yaml federate.sample_client_rate 1.0`

But 
`python federatedscope/main.py --cfg federatedscope/gfl/baseline/fedavg_gnn_node_fullbatch_citation.yaml federate.sample_client_num 5` works fine.

In fact, `federate.sample_client_rate=1.0` equals `federate.sample_client_num=5`. @yxdyc @rayrayraykk

PR #17, opened 2022-04-13
[Bugfix] fixed the invalid logger set-up if the logging is used before we call setup_logger

yxdyc:
fixed the invalid logger set-up if the `logging` is used before we call `setup_logger`; resolved the issue #12


rayrayraykk:
It looks good to me.

joneswong (approved):
looks good to me

PR #19, opened 2022-04-15
Communication efficiency optimization

xieyxclack:
1. The messages are automatically transformed according to their type (e.g., int, float, str) before being fed into the gRPC communicator, which reduces 20%-50% of the message sizes when running toy example in distributed mode. In the previous version, messages are transformed to json_str first and then fed into the gRPC communicator.  For more details please refer to:
   * A new proto file `gRPC_comm_manager.proto`;  
   * The `transform` function and `parse` function in `message.py`;  

2. Minor fix for running with distributed mode.


DavdGao:
It looks good to me. (Maybe we can add a unit test for the distributed mode?)

DavdGao:
It looks good to me. (Maybe we can add a unit test for the distributed mode?)

DavdGao (commented):
looks good to me

joneswong (approved):
The scope of supported types are any composition of dict/list/basic types (int, float, str). It is consistent with what currently being considered in our FL algos.

Approve!

Issue #18, opened 2022-04-14
Fail to run the standalone example with the minimal version of requirement 


xieyxclack:
An Error happens in `FederatedScope/federatedscope/cv/dataset/leaf_cv.py`:

When it is required to download the dataset, a function `download_url` is imported from the module `torch_geometric.data`, which is not included in the minimal version of the requirement, (of course it shouldn't be included in the minimal version, in my opinion).

So maybe we should replace the `download_url` with another implementation here. Thanks :)

Issue #16, opened 2022-04-13
Cannot import federatedscope


joneswong:
Although I have successfully completed the installation procedure, I cannot `import federatedscope` from path other than the root of the cloned repo.

PR #20, opened 2022-04-15
Change source of `download_url` from our own and fix `README`

rayrayraykk:
#18 #16 
As the title said.

### DIFF
--- a/federatedscope/core/auxiliaries/utils.py
+++ b/federatedscope/core/auxiliaries/utils.py
@@ -4,8 +4,10 @@ import os.path as osp
 import random
 import copy
 import sys
+import ssl
 import time
 import math
+import urllib.request
 from datetime import datetime
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


xieyxclack (dismissed):
Good job, it looks good to me and has resolved the proposed issue #18

yxdyc:
LGTM.

federatedscope/core/auxiliaries/utils.py
```diff
@@ -335,3 +337,31 @@ def merge_dict(dict1, dict2):
             else:
                 dict1[key].append(value)
     return dict1
+
+
+def download_url(url: str, folder='folder'):
```
DavdGao, line 342: Please check if there is other versions of the functions `download_url` and `extract`. (currently `mf` uses the torch version)

rayrayraykk: The minimal version `requirements-torch1.10.txt` contains `torchvision`,  so I think the usage of `download_url` from it is ok.

rayrayraykk:
> LGTM. One minor question, is the `download_url` copied from `torch_geometric.data`? If yes, we may add simple comment to clarify this @rayrayraykk.

Thanks, I have added comments.

joneswong (approved):
approve

Issue #22, opened 2022-04-18
Redundancy in the log files


rayrayraykk:
A Fedavg on 5% of FEMNIST trail will produce a 500 kb log each round:
with 80% eval logs like `2022-04-13 16:33:24,901 (client:264) INFO: Client #1: (Evaluation (test set) at Round #26) test_loss is 79.352451`. And 10% is server results and 10% is train informations. 

If the round is 500, 1000, or much larger, the log files will take up too much space with a lot of redundancy. @yxdyc

yxdyc:
The output logging files are currently redundant, leading to disk space wastage and difficulties to find key results from the logging files. We may save the ``raw_metrics'' and ``training_metrics'' in other two log files with compact format.

rayrayraykk:
> The output logging files are currently redundant, leading to disk space wastage and difficulties to find key results from the logging files. We may save the `raw_metrics` and `training_metrics` in other two log files with compact format.

And we should delete the logs below, which have already been reported on the server-side.
https://github.com/alibaba/FederatedScope/blob/24deff53055d309c3ef02e58cb7a544036b4d5ec/federatedscope/core/worker/client.py#L260

rayrayraykk:
Another issue is the eval results of the global evaluation mode are empty, we should fix this.

yxdyc:
> > The output logging files are currently redundant, leading to disk space wastage and difficulties to find key results from the logging files. We may save the `raw_metrics` and `training_metrics` in other two log files with compact format.
> 
> And we should delete the logs below, which have already been reported on the server-side.
> 
> https://github.com/alibaba/FederatedScope/blob/24deff53055d309c3ef02e58cb7a544036b4d5ec/federatedscope/core/worker/client.py#L260

We still need this log when we are in distributed mode. Do you agree we mask these logs only in the ``simulation'' mode? @rayrayraykk

rayrayraykk:
> > > The output logging files are currently redundant, leading to disk space wastage and difficulties to find key results from the logging files. We may save the `raw_metrics` and `training_metrics` in other two log files with compact format.
> > 
> > 
> > And we should delete the logs below, which have already been reported on the server-side.
> > https://github.com/alibaba/FederatedScope/blob/24deff53055d309c3ef02e58cb7a544036b4d5ec/federatedscope/core/worker/client.py#L260
> 
> We still need this log when we are in distributed mode. Do you agree we mask these logs only in the ``simulation'' mode? @rayrayraykk

[...]

PR #29, opened 2022-04-20
Refactored the logger by reducing its redundancy and fixed some minor issues

yxdyc:
Reducing the redundancy of the logger according to #22 . Specifically, the changes are:
- save the raw results in another file and compress it
- in standalone mode, by default, print the trainer meta-info once
- in standalone mode, do not print the local eval results from client-side
- fix the missing global eval results

Besides, fix some bugs in unit tests and outdir
- fix minor errors in the client id print and config yaml files
- fix the recursive outdir (#32) 
- unify the global_cfg usage in the unit tests (#34)


federatedscope/core/auxiliaries/utils.py
```diff
@@ -60,14 +58,15 @@ def setup_logger(cfg):
...
     # if exist, make directory with given name and time
     if os.path.isdir(cfg.outdir) and os.path.exists(cfg.outdir):
-        cfg.outdir = os.path.join(cfg.outdir, "sub_exp" +
-                                  datetime.now().strftime('_%Y%m%d%H%M%S')
-                                  )  # e.g., sub_exp_20220411030524
+        outdir = os.path.join(cfg.outdir, "sub_exp" +
+                              datetime.now().strftime('_%Y%m%d%H%M%S')
+                              )  # e.g., sub_exp_20220411030524
         while os.path.exists(cfg.outdir):
             time.sleep(1)
```
DavdGao, line 65: Why we need `time.sleep(1)` here?

federatedscope/core/auxiliaries/utils.py
```diff
@@ -60,14 +58,15 @@ def setup_logger(cfg):
...
-            cfg.outdir = os.path.join(
+            outdir = os.path.join(
                 cfg.outdir,
                 "sub_exp" + datetime.now().strftime('_%Y%m%d%H%M%S'))
+        cfg.outdir = outdir
```
DavdGao, line 69: It seems like `cfg.outdir` is initialized here by `exp/sub_exp__%Y%m%d%H%M%S`. Please make sure the config file is stored in the same dir (cfg is stored in the `freeze` function in config.py).

federatedscope/core/monitor.py
```diff
@@ -1,52 +1,168 @@
...
+        return round_formatted_results
+
+    def calc_blocal_dissim(last_model, local_updated_models):
+        '''
+        Arguments:
+            last_model (dict): the state of last round.
+            local_updated_models (list): each element is ooxx.
+        Returns:
+            b_local_dissimilarity (dict): the measurements.
+        '''
```
DavdGao, line 131: please elaborate the metric of dissimilarity.

DavdGao (commented):
please see the detailed comments.

federatedscope/core/auxiliaries/utils.py
yxdyc: for the case that in a very short time, some programs running in parallel create a directory with the same name

federatedscope/core/auxiliaries/utils.py
yxdyc: We will call the logger_setup at the beginning and the following usages are based on the same outdir, see the new [PR](https://github.com/alibaba/FederatedScope/pull/30/files#diff-eb363534ce61d0fc725e670c047ab83e5358698478d3065641d4b0b4445baff6).

federatedscope/core/monitor.py
yxdyc: Reference added

federatedscope/main.py
```diff
@@ -1,7 +1,7 @@
 import os
 import sys
 
-DEV_MODE = False  # simplify the federatedscope re-setup everytime we change the source codes of federatedscope
+DEV_MODE = True  # simplify the federatedscope re-setup everytime we change the source codes of federatedscope
```
rayrayraykk, line 4: We'd better keep it `False`.

tests/test_rec_IG_opt_attack.py
```diff
@@ -51,16 +51,17 @@ def set_config_femnist(self, cfg):
...
-        data, modified_cfg = get_data(global_cfg.clone())
-        global_cfg.merge_from_other_cfg(modified_cfg)
+        data, modified_cfg = get_data(init_cfg.clone())
+        init_cfg.merge_from_other_cfg(modified_cfg)
         self.assertIsNotNone(data)
 
         Fed_runner = FedRunner(data=data,
-                               server_class=get_server_cls(global_cfg),
+                               server_class=get_server_cls(init_cfg),
                                client_class=get_client_cls(global_cfg),
```
rayrayraykk, line 65: Should be `init_cfg`.

rayrayraykk (approved):

federatedscope/core/fed_runner.py
```diff
@@ -30,6 +30,10 @@ def __init__(self,
         if self.mode == 'standalone':
             self.shared_comm_queue = deque()
             self._setup_for_standalone()
+            # in standalone mode, by default, we print the trainer info only once for better logs readability
+            trainer_representative = self.client[1].trainer
```
joneswong, line 34: why is `client[1]`? any similar concept like "chief_worker" in ps-worker paradigm?

joneswong (approved):
approved. please have a look at my inline comments.

federatedscope/core/fed_runner.py
yxdyc: Here the "[1]" just indicates the first client since we use at least one clients in all cases. 

Arguably, we can modify the `trainer_representative ` access accordingly after we introducing a similar concept like "chief_worker" in future PR, which may be required by asynchronous simulation.

PR #36, opened 2022-04-22
Timeout strategy and minimal received number

xieyxclack:
[feature] For async: 
* Timeout strategy: to allow the server to stop waiting for the feedback when exceeding the given time budget; 
* Minimal received number/minimal received ratio: the minimal number of the received number for the server to move on;

Minor fix/modification;
* In distributed mode: handle the exception when grpc_channel is failed;
* Reorder the config to ensure the base config will be registered first;
* Clean the msg_buffer for eval;
* Dynamic track of the round number of the evaluation (in the previous version, the round number of the evaluation is set to be that of training minus one)
* Some logging;

TODO (for async, in another pr):
* handle the stale message


federatedscope/core/worker/server.py
```diff
@@ -178,9 +178,30 @@ def run(self):
...
+        min_received_num = self._cfg.asyn.min_received_num if hasattr(self._cfg, 'asyn') else self._cfg.federate.sample_client_num
+        with Timeout(self._cfg.asyn.timeout) as time_counter:
+            while self.state <= self.total_round_num:
+                try:
+                    msg = self.comm_manager.receive()
+                    move_on_flag = self.msg_handlers[msg.msg_type](msg)
+                    if move_on_flag:
+                        time_counter.reset()
+                except TimeoutError:
+                    print('Time out!')
```
DavdGao, line 190: Maybe we should print it out by `logger.info`.

federatedscope/core/worker/server.py
```diff
@@ -178,9 +178,30 @@ def run(self):
...
+                    move_on_flag_eval = self.check_and_move_on(min_received_num=min_received_num, check_eval_result=True)
+                    move_on_flag = self.check_and_move_on(min_received_num=min_received_num)
+                    if not move_on_flag and not move_on_flag_eval:
+                        ## Time out, broadcast the model para and re-start the training round
+                        logging.info(
+                            '----------- Re-starting the training round (Round #{:d}) -------------'
+                                .format(self.state))
+                        # Clean the msg_buffer
+                        self.msg_buffer['train'][self.state].clear()
```
DavdGao, line 199: Will the normal clients repeat local training and sending the same local parameters (since the global model and hyperparameters doesn't change) for the absence of some other clients?
Maybe we should set a maximum number of attempts, and report the failure of fl training after that.

DavdGao (commented):
please see the detailed comments

federatedscope/core/worker/server.py
xieyxclack: Here we restart the training round, including sample involved clients, local training, and then perform aggregation. So a client might repeat local training if it was sampled again.
We are going to implement a "cache" strategy to prevent redundant training, and it is a good idea to leave the failure exit here. Thanks!

joneswong (approved):
LGTM

PR #37, opened 2022-04-24
[Feature] FedEx

joneswong:
Adapted the open-sourced FedEx, a weight-sharing inspired fed hpo method, to our package.


federatedscope/autotune/fedex/client.py
```diff
@@ -0,0 +1,80 @@
...
+        for hyper in hyperparams:
+            for k, v in hyper.items():
+                cmd_args.append(k)
+                cmd_args.append(v)
+
+        self._cfg.defrost()
+        self._cfg.merge_from_list(cmd_args)
+        self._cfg.freeze()
+
+        self.trainer.ctx.setup_vars()
```
joneswong, line 28: @DavdGao hi dawei, please help me validate whether this line can actually update the hyper-parameters of the client. thx!

DavdGao: Executing the code in line 24-26 will change the attribute `self.trainer.ctx.cfg` and calling function `setup_vars` will update the hyper-parameters and partial variables. However, the variables like `model`, `data`, `device` in `self.trainer.ctx` won't change as they are already created within `self.trainer.ctx`.

federatedscope/core/configs/cfg_training.py
```diff
@@ -22,6 +22,7 @@ def extend_training_cfg(cfg):
     cfg.optimizer.type = 'SGD'
     cfg.optimizer.lr = 0.1
     cfg.optimizer.weight_decay = .0
+    cfg.optimizer.mmt = .0
```
rayrayraykk, line 25: I'm not sure the abbreviation for momentum is general enough.

rayrayraykk (commented):
Looks good! Please see the inline comments, thanks.

federatedscope/core/configs/cfg_training.py
joneswong: changed to "momentum" accordingly

DavdGao (commented):
looks good to me

rayrayraykk (approved):
Looks good to me.

PR #43, opened 2022-04-28
Added the local training baseline with clients' local early-stoppers

yxdyc:
- added the local training baseline
- enabled each client has its own early-stopper


federatedscope/core/monitors/monitor.py
```diff
@@ -176,3 +176,98 @@ def calc_blocal_dissim(last_model, local_updated_models):
             b_local_dissimilarity[k] = np.sqrt(
                 avg_gnorms[k].item() / torch.sum(global_grads[k]**2).item())
         return b_local_dissimilarity
+
+
+def update_best_result(best_results,
```
rayrayraykk, line 181: Please return `best_results`, or the value won't be changed.

rayrayraykk, line 181: Please return `best_results`, or the value won't be changed.

rayrayraykk (commented):
Please fix the bug, thanks!

DavdGao (approved):
looks good to me

PR #44, opened 2022-04-28
Rebuild and modify context accordingly

DavdGao:
- The attributes of `ctx` are classified into the following two classes
     - `ctx.xxx`: general attributes
     - `ctx.var.xxx`: actually refer to `ctx["var"][${ctx.cur_mode}_${ctx.cur_data_split}]["xxx"]`

 - Add lifecycle manager
     - Build two classes `CtxStatsVar` and `CtxReferVar` to wrap the variables with specific lifecycle and clear function. 
          - `CtxStatsVar`: statistic variables (float), e.g. `loss_batch_total`
               - `lifecycle`: chosen from  `batch`, `epoch`, `routine` and `None` 
          - `CtxReferVar`: reference variable, e.g. `y_prob`, `data_batch`
               - `lifecycle`: chosen from  `batch`, `epoch`, `routine` and `None` 
               - `efunc`: the customized function to run when the lifecycle ends
                   - e.g. for `ctx.model`, maybe we want to move it to cpu when the lifecycle ends. 
                   - When `efunc` is `None`, the variable is deleted at the end of the lifecycle
     - Add an attribute `lifecycles` in `Context` to record the lifecycles of the attributes
     - Add a decorator `@lifecycle(lifecycle=xxx)` to manage  the variables.


federatedscope/core/trainers/context.py
```diff
@@ -182,3 +217,116 @@ def reset_used_dataset(self):
...
+        super(CtxReferVar, self).__init__(lifecycle=lifecycle)
+        self.obj = obj
+        self.efunc = efunc
+
+    def clear(self):
+        if self.efunc is not None:
+            self.efunc(self.obj)
+
+
+def CtxStatsVar(init=0., lifecycle="routine"):
```
yxdyc, line 276: It seems that the `CtxStatsrVar` is a particular case of `CtxReferVar`, however, these two classes may confuse users to deciding which `XxxVar` for certain variables we need to use. Besides, the dynamic type specialization of `CtxStatsrVar` is a bit tricky. 
We can simply provide a `CtxVar` that contains any object with easy-to-use space-time management in terms of cur_mode/cur_dataset and lifecycles.

yxdyc (commented):
Looks good to me, please consider simplifying the different types of CtxVar (see inline comments)

DavdGao:
This pr is under modification

PR #47, opened 2022-04-29
[Feature] added global training mode via a proxy client that holds all data

yxdyc:
as the title says.


DavdGao:
looks good to me

DavdGao (approved):
looks good to me

PR #48, opened 2022-04-29
[Feature] added local fine-tuning before local evaluation

yxdyc:
As the title said and this PR is based on PR #47. 

One can firstly review and merge #47, then the modifications will be clearer for only the fine-tuning feature.


DavdGao (approved):
looks good to me

PR #52, opened 2022-05-04
added some core API reference

yxdyc:
As the title says. The pFL related comments have been within `federatedscope.core.trainers` in previous version.


joneswong (approved):
approved!

PR #66, opened 2022-05-06
Update federatedscope-torch1.10.Dockerfile

yezxxx:
remove nvidia apt repo source to avoid GPG error "public key not available“ under ubuntu 20.04


rayrayraykk:
Thanks for your PR, but please update your docker images to avoid this error:
`docker pull nvidia/cuda:11.3.1-runtime-ubuntu20.04`

This error is due to the Rotation of CUDA Linux Repository Key:
You can see it here: https://forums.developer.nvidia.com/t/notice-cuda-linux-repository-key-rotation/212772.

 And this has been fixed several days ago by NVIDIA, thanks again!

Issue #24, opened 2022-04-18
Monitoring improvement: visualization support 


yxdyc:
The monitoring can be improved in terms of the visualization support.

To use visualization tools such as wandb or tensorboard, we currently parse the log file after the results saved. We need to support results logging in real time rather the two-step style. Besides, the parsing process should be automatic for better usability.

We can discuss other requirements for visualization here @rayrayraykk

rayrayraykk:
* I'm not sure this feature will meet the demand of someone who attempts to get a learning curve with error bars (different seeds).
* We can group the same metric (with different splits) in one figure.

yxdyc:
Resolved in #90. If necessary, we can re-open this issue again.

Issue #23, opened 2022-04-18
Improve logging content with more metrics and better readability


yxdyc:
We aim to improve logging content with more metrics and better readability:

- the systematic metrics are missing. We need to add more metrics to reflect system performance, such as communication and computational efficiency. 

- we may add metric to reflect the convergence, e.g., the number of rounds to converge

We can discuss specific metric requirements and logging timing here @rayrayraykk

yxdyc:
Resolved in #90. If necessary, we can re-open this issue again.

PR #90, opened 2022-05-16
Enhanced monitor with system metrics logging and wandb support

yxdyc:
Enhanced monitor with system metrics logging and wandb support:

- supported monitoring several system-related metrics including flops, upload/download bytes, convergence rounds and time, according to #23 
- added wandb support, according to #24

### DIFF
--- a/federatedscope/core/monitors/monitor.py
+++ b/federatedscope/core/monitors/monitor.py
@@ -40,19 +40,19 @@ class Monitor(object):
         self.total_model_size = 0    # model size used in the worker, in terms of number of parameters
         self.flops_per_sample = 0    # average flops for forwarding each data sample
         self.flop_count = 0     # used to calculated the running mean for flops_per_sample
-        self.total_flops = -1  # total computation flops to convergence until current fl round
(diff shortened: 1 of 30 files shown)


federatedscope/core/monitors/monitor.py
```diff
@@ -16,18 +19,152 @@
...
+        if fl_mode == "standalone":
+            self.merge_system_metrics_simulation_mode()
+
+        if self.use_wandb:
+            try:
+                import wandb
+            except ImportError:
+                logger.error(
+                    "cfg.wandb.use=True but not install the wandb package")
+                exit()
```
rayrayraykk, line 147: How about raising ValueError?

rayrayraykk (commented):
Good job. And please see the inline comments.

joneswong (approved):
Approved! BTW, I noticed that several customized trainers have to explicitly register the newly added hook (seems to estimate flops). I am curious about when we should do that for a new trainer (e.g., the newly added FLIT by Haohui).

PR #98, opened 2022-05-20
Support to set different configs for different clients

DavdGao:
as the title says.

### DIFF
--- a/federatedscope/core/configs/config.py
+++ b/federatedscope/core/configs/config.py
@@ -48,16 +48,18 @@ class CN(CfgNode):
                         else:
                             del v[k]
 
-    def freeze(self):
+    def freeze(self, save=True):
         self.assert_cfg()
         self.clean_unused_sub_cfgs()
-        # save the final cfg
-        with open(os.path.join(self.outdir, "config.yaml"), 'w') as outfile:
-            from contextlib import redirect_stdout
-            with redirect_stdout(outfile):
-                tmp_cfg = copy.deepcopy(self)
-                tmp_cfg.cfg_check_funcs = []
-                print(tmp_cfg.dump())
+
+        if save:
+            # save the final cfg
+            with open(os.path.join(self.outdir, "config.yaml"), 'w') as outfile:
+                from contextlib import redirect_stdout
+                with redirect_stdout(outfile):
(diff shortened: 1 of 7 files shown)


federatedscope/main.py
```diff
@@ -34,8 +35,13 @@
 
     init_cfg.freeze()
 
+    # allow different settings for different clients
+    # cfg_client.merge_from_file(args.cfg_client)
+    cfg_client = CfgNode.load_cfg(open(args.cfg_client, 'r'))
```
xieyxclack, line 40: We should consider the situation that `args.cfg_client=None` or invalid path.

federatedscope/core/fed_runner.py
```diff
@@ -208,6 +210,12 @@ def _setup_client(self, client_id=-1, client_model=None):
 
         if self.client_class:
             client_specific_config = self.cfg.clone()
+            # Modify the setting according to self.cfg_client in standalone mode
+            if self.cfg_client is not None:
+                client_specific_config.defrost()
+                client_specific_config.merge_from_other_cfg(self.cfg_client.get('client_{}'.format(client_id)))
+                client_specific_config.freeze()
```
yxdyc, line 217: The `freeze()` will override the cfg.yaml file. We may need to add a parameter for this function to skip the save process

DavdGao: It is modifed accordingly.

federatedscope/main.py
DavdGao: It is modifed accordingly

federatedscope/core/configs/config.py
```diff
@@ -48,16 +48,18 @@ def clean_unused_sub_cfgs(self):
...
-            with redirect_stdout(outfile):
-                tmp_cfg = copy.deepcopy(self)
-                tmp_cfg.cfg_check_funcs = []
-                print(tmp_cfg.dump())
+
+        if save:
+            # save the final cfg
+            with open(os.path.join(self.outdir, "config.yaml"), 'w') as outfile:
+                from contextlib import redirect_stdout
+                with redirect_stdout(outfile):
```
joneswong, line 59: i am curious about why writing to a file by redirecting stdout.

joneswong (approved):
LGTM

PR #100, opened 2022-05-20
[Feature] Enabled un-seen clients case to check the participation generalization gap

yxdyc:
Enabled un-seen clients simulation, such that we can check the participation generalization gap, refer [ICLR'22, What Do We Mean by Generalization in Federated Learning?](https://openreview.net/forum?id=VimqQq-i_Q)

This PR is based on the another PR (enhanced monitor #90)

### DIFF
--- a/federatedscope/core/fed_runner.py
+++ b/federatedscope/core/fed_runner.py
@@ -2,6 +2,8 @@ import logging
 
 from collections import deque
 
+import numpy as np
+
 from federatedscope.core.worker import Server, Client
 from federatedscope.core.gpu_manager import GPUManager
 from federatedscope.core.auxiliaries.model_builder import get_model
@@ -89,7 +91,18 @@ class FedRunner(object):
                 self.cfg.federate.sample_client_num = 1
                 self.cfg.freeze()
 
+        unseen_clients_id = []
+        if self.cfg.federate.unseen_clients_rate > 0:
+            unseen_clients_id = np.random.choice(
+                np.arange(1, self.cfg.federate.client_num + 1),
+                size=max(
+                    1,
+                    int(self.cfg.federate.unseen_clients_rate *
+                        self.cfg.federate.client_num)),
+                replace=False).tolist()
+
         self.server = self._setup_server()
+        self.server.unseen_clients_id = unseen_clients_id
 
         self.client = dict()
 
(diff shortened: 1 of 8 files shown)


federatedscope/core/fed_runner.py
```diff
@@ -89,7 +91,18 @@ def _setup_for_standalone(self):
...
+            unseen_clients_id = np.random.choice(
+                np.arange(1, self.cfg.federate.client_num + 1),
+                size=max(
+                    1,
+                    int(self.cfg.federate.unseen_clients_rate *
+                        self.cfg.federate.client_num)),
+                replace=False).tolist()
+
         self.server = self._setup_server()
+        self.server.unseen_clients_id = unseen_clients_id
```
xieyxclack, line 105: IMO, we have better not modify the attributes of server/client in fed_runner

federatedscope/core/fed_runner.py
```diff
@@ -102,6 +115,8 @@ def _setup_for_standalone(self):
         for client_id in range(1, self.cfg.federate.client_num + 1):
             self.client[client_id] = self._setup_client(
                 client_id=client_id, client_model=self._shared_client_model)
+            if client_id in unseen_clients_id:
+                self.client[client_id].is_unseen_client = True
```
xieyxclack, line 119: Like above

federatedscope/core/sampler.py
```diff
@@ -23,14 +23,19 @@ def change_state(self, indices, state):
         """
         To modify the state of clients (idle or working)
         """
-        if state == 'idle':
-            self.client_state[indices] = 1
-        elif state == 'working':
-            self.client_state[indices] = 0
+        if not isinstance(indices, list):
```
xieyxclack, line 26: Shall we cover np.array here?

xieyxclack (commented):
LGTM, please see the inline comments for minor suggestions. And I am not sure that whether setting `unseen_clients` would affect the behaviors of broadcasting to all neighbors in some applications.

federatedscope/core/fed_runner.py
yxdyc: moved it into the setup_server() as sub action of init

federatedscope/core/fed_runner.py
yxdyc: moved it into the setup_server() as sub action of init

federatedscope/core/sampler.py
yxdyc: modified accordingly, besides, we used another state named "unseen"

xieyxclack (approved):
LGTM

PR #118, opened 2022-05-26
Fscontest

wanghh7:
Update trainer and monitor for multi regression tasks.


joneswong:
these changes are expected to be included by our master branch in the future.

rayrayraykk (commented):
Looks good to me, but how about making [node-level](https://github.com/alibaba/FederatedScope/blob/167c1304d4f2e72511f421f330cf19113d123a1a/federatedscope/gfl/model/model_builder.py#L29) and [link-level](https://github.com/alibaba/FederatedScope/blob/167c1304d4f2e72511f421f330cf19113d123a1a/federatedscope/gfl/model/model_builder.py#L65) constant with graph-level tasks?

PR #119, opened 2022-05-26
[HotFix] Fix contrib.

rayrayraykk:
As the title says, could not use external file in `contrib`.

### DIFF
--- a/federatedscope/core/auxiliaries/data_builder.py
+++ b/federatedscope/core/auxiliaries/data_builder.py
@@ -4,6 +4,10 @@ from collections import defaultdict
 
 import federatedscope.register as register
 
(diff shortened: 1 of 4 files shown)


joneswong (commented):
looks good. why do we need try-catch here?

rayrayraykk:
> looks good. why do we need try-catch here?

We may add more examples, such as data from PYG, to the `contrib` in the future, but some examples are not available for users who have the minimal version installed.

joneswong:
sounds reasonable...but why not let people know what happened. For example, if the failure of importing is caused by my incorrect implementation.

joneswong (approved):
approved.

PR #120, opened 2022-05-26
Calibrate FedEx

joneswong:
In the original implementation of FedEx, the policy is allowed to use after - before as reward, where after and before refer to the validation loss estimated before after and before the local updates.

This pr extends the base torch trainer so that we can optionally conduct evaluation before and after the local updates.


joneswong:
@DavdGao @rayrayraykk As the torch trainer is moved to a new .py file in this pr, all the code is regarded as changes. Please focus on the `train()` method, where the actual changes happen.

DavdGao (commented):
Looks good to me

rayrayraykk (approved):
Looks good to me.

Issue #122, opened 2022-05-29
AttributeError: Attempted to set type to proximal_regularizer, but CfgNode is immutable.Could please help me to overcome this problem,which happens when I call for the fedprox algorithm:python federatedscope/main.py --cfg federatedscope/example_configs/femnist.yaml fedprox.use True


Zst0514:

DavdGao:
Thanks for your trying. It seems like the `cfg` in unchangable, your can try the following code in `federatedscope/core/trainers/trainer_nbafl.py`
```
    cfg.regularizer.defrost()
    cfg.regularizer.type = 'proximal_regularizer'
    cfg.regularizer.mu = cfg.fedprox.mu
    cfg.regularizer.freeze(save=False)
```
, and we will fix this problem quickly.

xieyxclack:
This thread is closed as we fix this problem in #124 
Thanks again for your comments!

Zst0514:
I was a little busy the other day, sorry for replying to your email just now, thank you very much for your answer!I may ask you for more questions related in the future.













At 2022-06-02 11:45:32, "Yuexiang XIE" ***@***.***> wrote:

Closed #122 as completed.

PR #124, opened 2022-05-30
[Hotfix] Hotfix for issue #122

DavdGao:
As the title says


yxcdy:
LGTM

yxdyc:
LGTM.

federatedscope/core/trainers/trainer_nbafl.py
```diff
@@ -54,8 +54,10 @@ def init_nbafl_ctx(base_trainer):
     cfg = base_trainer.cfg
 
     # set proximal regularizer
+    cfg.regularizer.defrost()
     cfg.regularizer.type = 'proximal_regularizer'
     cfg.regularizer.mu = cfg.nbafl.mu
+    cfg.regularizer.freeze(save=False)
```
joneswong, line 60: when should we set `save` to be `False`?

joneswong (dismissed):
looks good. plz resolve the conflicts. just a kind reminder: whether FedEx client needs to set `save` to be `False`

federatedscope/core/trainers/trainer_nbafl.py
DavdGao: In my opinion, the target of saving configuration is to reproduct the experiment. So maybe we should save the `config`  before calling `FedRunner().run()`

DavdGao:
> looks good. plz resolve the conflicts. just a kind reminder: whether FedEx client needs to set `save` to be `False`

Thanks, the conflicts are resolved.

federatedscope/core/trainers/trainer_nbafl.py
xieyxclack: But the config might be changed to "valid" value in some assertions and saved. I think the saved config implies what takes effect during the exp.

DavdGao: > 

I have checked the code, and delete the `save` parameter in `cfg.freeze()`.

xieyxclack (approved):
It looks good to me. Thanks a lot @DavdGao

PR #145, opened 2022-06-06
Improve SHA wrapper

joneswong:
enable SHA wrapper to distinguish hyperparameters can/cannot be handled by FedEx and generate the arms for each FedEx via local perturbation.


rayrayraykk (approved):
LGTM

PR #149, opened 2022-06-10
init the pFL-bench scripts

yxdyc:
init the pFL-bench scripts and README.


DavdGao:
> 

I'm not sure, maybe it's better to place the scripts in the same directory for both master and Feature/pfl_bench?

yxdyc:
> > 
> 
> I'm not sure, maybe it's better to place the scripts in the same directory for both master and Feature/pfl_bench?

If in master, this PR will involves ~600 files and 5w+ codes. So I put them into a zip file, and refer readers to the file link as README said: `You can find the full scripts from the another [branch](https://github.com/alibaba/FederatedScope/tree/Feature/pfl_bench/scripts/personalization_exp_scripts/pfl_bench) or the packed small zip [file](https://federatedscope.oss-cn-beijing.aliyuncs.com/pfl_bench_scripts.zip), in which we organize the scripts for all the methods and all the datasets as multiple directories named by dataset name.`

xieyxclack (approved):
LGTM

PR #154, opened 2022-06-13
update splitter for consistent label distribution

xieyxclack:
The splitters are allowed to partition the data into multiple clients according to the given label distribution


DavdGao (approved):
looks good to me

DavdGao (approved):

PR #162, opened 2022-06-14
Update baseline codes for FedNLP

cheneydon:
As the title says.


federatedscope/core/auxiliaries/trainer_builder.py
```diff
@@ -122,7 +126,11 @@ def get_trainer(model=None,
 
     # fed algorithm plug-in
     if config.fedprox.use:
-        from federatedscope.core.trainers.trainer_fedprox import wrap_fedprox_trainer
-        trainer = wrap_fedprox_trainer(trainer)
+        if config.federate.method.lower() == "fedprox":
```
DavdGao, line 129: We should consider the scenarios that `fedprox.use=True` but `federate.method='fedem'/'ditto'/'fedbn'/...`

federatedscope/main.py
```diff
@@ -10,37 +10,36 @@
...
 from federatedscope.core.fed_runner import FedRunner
-from yacs.config import CfgNode
+from federatedscope.nlp.configs.cfg_client import extend_cfg_client
 
 if os.environ.get('https_proxy'):
     del os.environ['https_proxy']
 if os.environ.get('http_proxy'):
     del os.environ['http_proxy']
 
+os.environ['TOKENIZERS_PARALLELISM'] = 'false'
```
DavdGao, line 22: what is this `TOKENIZERS_PARALLELISM` for?

federatedscope/main.py
```diff
@@ -10,37 +10,36 @@
...
-
     init_cfg.freeze()
 
     # allow different settings for different clients
-    # cfg_client.merge_from_file(args.cfg_client)
     if args.cfg_client is None:
         cfg_client = None
     else:
-        cfg_client = CfgNode.load_cfg(open(args.cfg_client, 'r'))
+        cfg_client = CN.load_cfg(open(args.cfg_client, 'r'))
```
DavdGao, line 41: The dataset Graph-DT also uses `args.cfg_client`, which may cause certain error here. 
please see the comments in `federatedscope/nlp/configs/cfg_client.py`

federatedscope/nlp/configs/cfg_client.py
```diff
@@ -0,0 +1,19 @@
+import os.path as osp
+import copy
+
+
+def extend_cfg_client(init_cfg, cfg_client):
+    num_clients = len([k for k in cfg_client.keys() if k.startswith('client')])
+    for i in range(1, num_clients + 1):
+        cfg = cfg_client['client_{}'.format(i)]
+        task = cfg.data.type
+        cfg.data.batch_size = init_cfg.data.all_batch_size[task]
```
DavdGao, line 10: It seems like your want to set the `batch_size` according to the type of tasks. 
For now, the type of task in each client is decided by the dataloader, which means we know already know the task type of each client. Therefore, we can set the cfg_client as follows to set different parameters. 
```
client_1:
    data:
        batch_size: 20
client_2:
    data:
        batch_size: 30
client_3:
    data:
        batch_size: 50
```

federatedscope/nlp/trainer/fedprox_trainer.py
```diff
@@ -0,0 +1,74 @@
+import logging
```
DavdGao, line 1: it seems like there is no difference with the original `core/trainer/fedprox_trainer.py`. maybe we can reuse the one in core/trainer?

DavdGao (commented):

DavdGao (approved):
Good work, it looks good to me.

Issue #132, opened 2022-06-01
Invalid system metric names


joneswong:
There are metrics that have invalid names. It seems that this is caused by incorrect recursive concatenation of strings.

PR #175, opened 2022-06-20
Bugfix for the invalid system metric names, via using worker-specific monitor

yxdyc:
Bugfix for the invalid system metric names, via using worker-specific monitor (issue #132). 

(Note: the main changes are for monitor.py, server.py and client.py;  other files are changed due to the linter)


xieyxclack (approved):
LGTM

DavdGao (approved):
looks good to me, and maybe we should update the way of naming in our tutorial

PR #185, opened 2022-06-23
[WIP] merge the test set for evaluation (for efficient simulation)

xieyxclack:
As the title says.


federatedscope/core/configs/cfg_fl_setting.py
```diff
@@ -24,6 +24,7 @@ def extend_fl_setting_cfg(cfg):
     cfg.federate.online_aggr = False
     cfg.federate.make_global_eval = False
     cfg.federate.use_diff = False
+    cfg.federate.merge_eval = False
```
DavdGao, line 27: It should be `merge_test` here.

federatedscope/core/configs/cfg_fl_setting.py
```diff
@@ -134,5 +135,9 @@ def assert_fl_setting_cfg(cfg):
         not cfg.federate.use_ss
     ), "Have not supported to use online aggregator and secrete sharing at the same time"
 
+    if cfg.federate.merge_eval and not cfg.federate.make_global_eval:
+        cfg.federate.make_global_eval = True
+        logger.warning('Set cfg.federate.make_global_eval=True since cfg.federate.merge_eval=True')
+
```
DavdGao, line 141: The same as above.

federatedscope/core/fed_runner.py
```diff
@@ -156,7 +156,13 @@ def _setup_server(self):
         """
         self.server_id = 0
         if self.mode == 'standalone':
-            if self.server_id in self.data:
+            if self.cfg.federate.merge_test:
+                from federatedscope.core.auxiliaries.data_builder import merge_test_data
+                num_of_sample, server_data = merge_test_data(all_data=self.data)
+                model = get_model(self.cfg.model,
+                              server_data,
```
DavdGao, line 163: It will report error when initialize a trainer within the server, since the `get_cnn` function requires some data from 'train' (federatedscope/cv/model/model_builder.py) to initialize the CNN models but the server only has 'test' data.

DavdGao: Also, the number of training data (`ctx.num_train_data=0`) in the server will cause some error for dp algorithm [`NBAFL` ](https://github.com/alibaba/FederatedScope/blob/master/federatedscope/core/trainers/trainer_nbafl.py#L65).

DavdGao (commented):

DavdGao:
Besides, if we choose `dirichlet_distribution_noniid_slice`, it seems like we must set `data.consistent_label_distribution` as `True`, otherwise the program cannot finish the partition.

xieyxclack:
This thread is closed since it has been merged into #225

PR #200, opened 2022-07-05
[feature] samplers

xieyxclack:
The sampler is used to sample clients from the idle clients at each training round

### DIFF
--- a/federatedscope/core/sampler.py
+++ b/federatedscope/core/sampler.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 
 import numpy as np
 
+
 class Sampler(ABC):
     """
     The strategies of sampling clients for each training round
@@ -10,7 +11,7 @@ class Sampler(ABC):
         client_state: a dict to manager the state of clients (idle or busy)
     """
     def __init__(self, client_num):
-        self.client_state = np.asarray([1]*(client_num+1))
+        self.client_state = np.asarray([1] * (client_num + 1))
         # Set the state of server (index=0) to 'working'
         self.client_state[0] = 0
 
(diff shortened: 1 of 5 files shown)


federatedscope/core/sampler.py
```diff
@@ -0,0 +1,124 @@
...
+        super(UniformSampler, self).__init__(client_num)
+
+    def sample(self, size):
+        """
+        To sample clients
+        """
+        idle_clients = np.nonzero(self.client_state)[0]
+        sampled_clients = np.random.choice(idle_clients,
+                                           size=size,
+                                           replace=False).tolist()
```
joneswong, line 50: Is it possible to make this argument configurable?

federatedscope/core/sampler.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    The strategies of sampling clients for each training round
+
+    Arguments:
+        client_state: a dict to manager the state of clients (idle or busy)
+    """
+    def __init__(self, client_num):
+        self.client_state = np.asarray([1] * (client_num + 1))
+        # Set the state of server (index=0) to 'working'
+        self.client_state[0] = 0
```
joneswong, line 16: it is very confusing. Server state is maintained in  an array named "client_state"

xieyxclack: Since 'client_id' is numbered from 1 to M, `index=0` for `client_state` is redundant. Although `client_state[0]` is used for `server_state`, it would not be modified during the training process.
Maybe we can declare it in the comments.

federatedscope/core/sampler.py
xieyxclack: Do you mean `replace`? It should be set to `False` to avoid sampling repeated clients.

PR #204, opened 2022-07-05
Support more datasets for distribute mode

xieyxclack:
As the title says. Users can specify `distribute.data_idx` to represent which partition of the data is used for the participant, (it would be randomly sampled if an invalid data_idx is provided).


joneswong (approved):
LGTM

PR #205, opened 2022-07-05
Changes for format checking

rayrayraykk:
1. Upper the lower bound for pyroma
2. Improve README

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ python setup.py install
 
 # Or (for dev mode)
 pip install -e .[dev]
+pre-commit install
 ```
 
 ##### From conda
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -82,6 +82,7 @@ python setup.py install
 
 # Or (for dev mode)
 pip install -e .[dev]
+pre-commit install
```
joneswong, line 85: I hv no idea about this. is it related to installation?

joneswong (approved):
approved!

README.md
rayrayraykk: For dev mode, if `pre-commit` is not installed in the repo, the checks won't work like `yapf`, `flake8`.

Issue #198, opened 2022-06-30
AttributeError in distributed mode -- (Avoid type conversion outside worker)


rayrayraykk:
**Describe the bug**
When model contains BN layer, the param `bn.num_batches_tracked` would be convert to `int` by `grpc`. But the `trainer.update` can't handle this situation well.

![image](https://user-images.githubusercontent.com/39145382/176586086-6a99e48d-4e1a-4758-9920-077e6681c5c3.png)

A dummy solution:
```python
  def update(self, model_parameters):
      '''
          Called by the FL client to update the model parameters
      Arguments:
          model_parameters (dict): PyTorch Module object's state_dict.
      '''
      for key in model_parameters:
          if isinstance(model_parameters[key], list):
              model_parameters[key] = torch.FloatTensor(
                  model_parameters[key])
          elif isinstance(model_parameters[key], int):
              model_parameters[key] = torch.tensor(model_parameters[key], dtype=torch.long)
              print(key, model_parameters[key])
          elif isinstance(model_parameters[key], float):
              model_parameters[key] = torch.tensor(model_parameters[key], dtype=torch.float)
      self.ctx.model.load_state_dict(self._param_filter(model_parameters),
                                     strict=False)
```

or can we solve it before sending the model_param?

rayrayraykk:
There is another type conversion bug in [aggregator](https://github.com/alibaba/FederatedScope/blob/2c252e72c78e5f0ef2112779a5098b74b2be223d/federatedscope/core/aggregator.py#L86).

If the value is int, like 8, FloatTensor(8), would convert 8 to a random tensor with shape 8.
Should we handle these two situation in the message buffer instead of in `aggregator`?

Need a discuss @xieyxclack @joneswong .

A dummy solution in aggregator:
```python
  if not self.cfg.federate.use_ss:
      if isinstance(local_model[key], torch.Tensor):
          local_model[key] = local_model[key].float()
      elif isinstance(local_model[key], list):
          local_model[key] = torch.FloatTensor(local_model[key])
      elif isinstance(local_model[key], int):
          local_model[key] = torch.tensor(local_model[key], dtype=torch.long)
      elif isinstance(local_model[key], float):
          local_model[key] = torch.tensor(local_model[key], dtype=torch.float)
```

xieyxclack:
> Avoid type conversion outside worker

IMO, the type conversion cannot be happened inside the worker, since the worker should be language-independent

PR #213, opened 2022-07-07
[HotFix] workaround for type convertion

rayrayraykk:
As issue #198 discussed, we use this workaround for the cloud demo.


joneswong (approved):
approved!

PR #215, opened 2022-07-08
support logging client-wise training info via wandb

yxdyc:
As the title says. The modification has been reviewed and merged in Master.


rayrayraykk (approved):
LGTM

PR #220, opened 2022-07-08
support both pre- and post-merging data for the "global" baseline

yxdyc:
as the title says

### DIFF
--- a/federatedscope/core/configs/cfg_data.py
+++ b/federatedscope/core/configs/cfg_data.py
@@ -23,7 +23,8 @@ def extend_data_cfg(cfg):
     cfg.data.drop_last = False
     cfg.data.sizes = [10, 5]
     cfg.data.shuffle = True
-    cfg.data.server_holds_all = False  # whether the server (client[0]) holds all data, useful in global evaluation case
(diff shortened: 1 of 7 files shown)


rayrayraykk (dismissed):
LGTM, but should we provide some baseline yaml files?

federatedscope/core/configs/cfg_data.py
```diff
@@ -23,6 +23,7 @@ def extend_data_cfg(cfg):
     cfg.data.drop_last = False
     cfg.data.sizes = [10, 5]
     cfg.data.shuffle = True
+    cfg.data.server_holds_all = False  # whether the server (client[0]) holds all data, useful in global evaluation case
```
xieyxclack, line 26: It is confusing to claim server = client[0]

xieyxclack (commented):
LGTM

federatedscope/core/configs/cfg_data.py
yxdyc: modified accordingly

yxdyc:
> LGTM, but should we provide some baseline yaml files?

added accordingly

xieyxclack (approved):

Issue #108, opened 2022-05-24
Terse logging


joneswong:
The results (most are floating numbers) are printed to the stdout without controlling the preserved precision. Thus, the reported results look lengthy.

PR #226, opened 2022-07-13
enabled precision configuartion when printing out float numbers

yxdyc:
as the title says, to address the issue #108.


xieyxclack (approved):
LGTM

PR #236, opened 2022-07-14
Construct FL course when server does not have data

### LINKED ISSUE #208: Support run FL courses when the server doesn't have data
In the current version, we need the `server_data` for building a model and getting a trainer, but in some cases, the server doesn't have the data or only has test data. 
Furthermore, does the server really need a trainer?  I find that the server's trainer is only used for evaluation.

xieyxclack:
As the title says. When the server does not have data:
- if some data is available as data representative (e.g., in simulation the client data), these data can be used to extract the input shape for building models at the server;
- Otherwise, the users can specify `model.input_shape` to tell what's the input shape (a tuple) to get the model if necessary;
- fix #208 

When the server does not have data in simulation, users can set `data.type=''`. (Please see the examples in scripts/run_distributed_lr.sh)

Please @rayrayraykk check `get_shape_from_data` in federatescope/core/auxiliaries/model_builder.py, thx!

### DIFF
--- a/federatedscope/core/auxiliaries/data_builder.py
+++ b/federatedscope/core/auxiliaries/data_builder.py
@@ -575,6 +575,10 @@ def get_data(config):
     elif 'cikmcup' in config.data.type.lower():
         from federatedscope.gfl.dataset.cikm_cup import load_cikmcup_data
         data, modified_config = load_cikmcup_data(config)
+    elif config.data.type is None or config.data.type == "":
+        # The participant (only for server in this version) does not own data
+        data = None
+        modified_config = config
     else:
         raise ValueError('Data {} not found.'.format(config.data.type))
 
@@ -582,7 +586,9 @@ def get_data(config):
         return data, modified_config
     else:
         # Invalid data_idx
-        if config.distribute.data_idx not in data.keys():
+        if config.distribute.data_idx == -1:
+            return data, config
+        elif config.distribute.data_idx not in data.keys():
             data_idx = np.random.choice(list(data.keys()))
             logger.warning(
                 f"The provided data_idx={config.distribute.data_idx} is "
--- a/federatedscope/core/auxiliaries/model_builder.py
+++ b/federatedscope/core/auxiliaries/model_builder.py
@@ -11,7 +11,76 @@ except ImportError as error:
         f'available.')
 
 
-def get_model(model_config, local_data, backend='torch'):
+def get_shape_from_data(data, model_config, backend='torch'):
+    """
+    Extract the input shape from the given data, which can be used to build
+    the data. Users can also use `data.input_shape` to specify the shape
+    Arguments:
+        data (object): the data used for local training or evaluation
+        The expected data format:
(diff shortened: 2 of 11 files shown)


federatedscope/core/auxiliaries/data_builder.py
```diff
@@ -572,14 +572,20 @@ def get_data(config):
...
+        modified_config = config
     else:
         raise ValueError('Data {} not found.'.format(config.data.type))
 
     if config.federate.mode.lower() == 'standalone':
         return data, modified_config
     else:
         # Invalid data_idx
-        if config.distribute.data_idx not in data.keys():
+        if config.distribute.data_idx == -1:
```
rayrayraykk, line 586: When `data.type` is `None` or `""` while `distribute.data_idx` is given, we should assert in  `configs`.

federatedscope/core/configs/cfg_fl_setting.py
```diff
@@ -49,7 +49,12 @@ def extend_fl_setting_cfg(cfg):
     cfg.distribute.client_port = 50050
     cfg.distribute.role = 'client'
     cfg.distribute.data_file = 'data'
```
rayrayraykk, line 51: The data-related keyword `distribute.data_file` in distributed mode seems to be the same meaning as `data.root`.

rayrayraykk (commented):
LGTM, but please see the inline comments, thanks.

federatedscope/core/auxiliaries/model_builder.py
```diff
@@ -11,99 +11,138 @@
...
+        1): {train/val/test: {x:ndarray, y:ndarray}}}
+        2): {train/val/test: DataLoader}
+    Returns:
+        shape (tuple): the input shape
+    """
+    # Handle some special cases
+    if model_config.type.lower() in ['vmfnet', 'hmfnet']:
+        return data['train'].n_col if model_config.type.lower(
+        ) == 'vmfnet' else data['train'].n_row
+    elif model_config.type.lower() in [
```
rayrayraykk, line 30: For graph-level tasks, data is  `{train/val/test: DataLoader}`.

federatedscope/core/auxiliaries/data_builder.py
xieyxclack: When `date.type` is None or "", the `distribute.data_idx` does not work. IMO, we can provide a WARNING but not an assertion here.

federatedscope/core/configs/cfg_fl_setting.py
xieyxclack: ‘data.root’ is a file directory and `distribute.data_file` includes the filename here

federatedscope/core/auxiliaries/model_builder.py
xieyxclack: I have fixed this issue, thx!

yxdyc (approved):
Good Job. LGTM.

PR #239, opened 2022-07-17
Support different types of key when communicating via grpc

xieyxclack:
As the title says. The kye types can be `int` or `string`.

### DIFF
--- a/federatedscope/core/communication.py
+++ b/federatedscope/core/communication.py
@@ -84,8 +84,14 @@ class gRPCCommManager(object):
         return server
 
     def add_neighbors(self, neighbor_id, address):
-        self.neighbors[neighbor_id] = '{}:{}'.format(address['host'],
-                                                     address['port'])
+        if isinstance(address, dict):
+            self.neighbors[neighbor_id] = '{}:{}'.format(
+                address['host'], address['port'])
+        elif isinstance(address, str):
+            self.neighbors[neighbor_id] = address
+        else:
+            raise TypeError(f"The type of address ({type(address)}) is not "
+                            "supported yet")
 
     def get_neighbors(self, neighbor_id=None):
         address = dict()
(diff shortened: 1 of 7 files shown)


federatedscope/core/message.py
```diff
@@ -180,7 +189,11 @@ def build_msg_value(self, value):
         if isinstance(value, list) or isinstance(value, tuple):
             msg_value.list_msg.MergeFrom(self.create_by_type(value))
         elif isinstance(value, dict):
-            msg_value.dict_msg.MergeFrom(self.create_by_type(value))
+            if isinstance(list(value.keys())[0], str):
```
rayrayraykk, line 192: What if a `dict` contains string key and int key?

rayrayraykk (commented):
Please see the inline comments, and the hotfix looks good to me.

federatedscope/core/communication.py
```diff
@@ -84,8 +84,14 @@ def serve(self, max_workers, host, port, options):
         return server
 
     def add_neighbors(self, neighbor_id, address):
-        self.neighbors[neighbor_id] = '{}:{}'.format(address['host'],
-                                                     address['port'])
+        if isinstance(address, dict):
+            self.neighbors[neighbor_id] = '{}:{}'.format(
+                address['host'], address['port'])
+        elif isinstance(address, str):
```
joneswong, line 90: We'd better have an assertion about the format of string

federatedscope/core/message.py
xieyxclack: Such situation cannot be handled due to the implementation grpc. A workaround can be that we change all the keys into string and save their original types for recovery, which might bring additional cost.

rayrayraykk: IMO, we'd better have a format check with dict values instead of using the first key.

joneswong (approved):
approved

PR #248, opened 2022-07-19
[Feature] Add learning rate scheduler in FS

DavdGao:
As the title says, add the support for learning rate scheduler

### DIFF
--- a/federatedscope/core/configs/cfg_training.py
+++ b/federatedscope/core/configs/cfg_training.py
@@ -22,9 +22,9 @@ def extend_training_cfg(cfg):
     cfg.train.optimizer.type = 'SGD'
     cfg.train.optimizer.lr = 0.1
 
+    # you can add new arguments 'aa' by `cfg.train.scheduler.aa = 'bb'`
     cfg.train.scheduler = CN(new_allowed=True)
     cfg.train.scheduler.type = ''
-    cfg.trainer.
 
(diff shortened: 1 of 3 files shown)


federatedscope/core/configs/cfg_training.py
```diff
@@ -22,6 +22,10 @@ def extend_training_cfg(cfg):
     cfg.train.optimizer.type = 'SGD'
     cfg.train.optimizer.lr = 0.1
 
+    # you can add new arguments 'aa' by `cfg.train.scheduler.aa = 'bb'`
+    cfg.train.scheduler = CN(new_allowed=True)
```
xieyxclack, line 26: Why we need to set `new_allowed=True` here?

xieyxclack (commented):
LGTM

federatedscope/core/configs/cfg_training.py
DavdGao: > Why we need to set `new_allowed=True` here?

It allow the users set different lr schedulers, and we don't need to list all the potential parameters. For example, if the type of lr scheduler is `StepLR`(torch.optim.lr_scheduler.StepLR), users can add their parameters as follows 
```
train:
    scheduler:
        type: StepLR
        step_size: 5
        gamma: 0.1
```

xieyxclack (approved):

PR #257, opened 2022-07-22
[Feature] Add organizer to FS

rayrayraykk:
# Organizer

Organizer is implemented by [Celery](https://docs.celeryq.dev/en/latest/) and [Redis](https://redis.io/) as Broker, so you should install it before using.

## Installation

```bash
# For Server
pip install -e .[org]
docker run -d -p 6379:6379 redis

# For Client
pip install -e .[org]
# Not full login ssh, will conduct `source ~/.bashrc` first.
# ** If not running interactively, `source ~/.bashrc` might fail **
# ** due to: `[ -z "$PS1" ] && return`, please comment this line **
```

## RUN

```bash
# For Server
cd federatedscope/organizer
celery -A server worker --loglevel=info

## For multi-worker
# celery multi start w1 -A server -l info
# celery multi start w2 -A server -l info
# ...

# For Client
# Modify `server_ip` in federatedscope/organizer/cfg_client.py
python federatedscope/organizer/client.py
help
```

### DIFF
--- a/federatedscope/core/configs/config.py
+++ b/federatedscope/core/configs/config.py
@@ -261,8 +261,8 @@ class CN(CfgNode):
                             "package")
                         exit()
 
-        if inform and save:
-            logger.info("the used configs are: \n" + str(tmp_cfg))
+            if inform:
+                logger.info("the used configs are: \n" + str(tmp_cfg))
 
         super(CN, self).freeze()
 
--- a/federatedscope/core/configs/config.py
+++ b/federatedscope/core/configs/config.py
@@ -112,7 +112,6 @@ class CN(CfgNode):
         containers to allow simple key-value access and management.
 
     """
-
     def __init__(self, init_dict=None, key_list=None, new_allowed=False):
         super().__init__(init_dict, key_list, new_allowed)
         self.__dict__["cfg_check_funcs"] = list(
@@ -258,10 +257,11 @@ class CN(CfgNode):
                         wandb.config.update(cfg_yaml, allow_val_change=True)
                     except ImportError:
                         logger.error(
-                            "cfg.wandb.use=True but not install the wandb package")
+                            "cfg.wandb.use=True but not install the wandb "
+                            "package")
                         exit()
 
-        if inform:
+        if inform and save:
             logger.info("the used configs are: \n" + str(tmp_cfg))
 
         super(CN, self).freeze()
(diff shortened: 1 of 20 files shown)


federatedscope/organizer/server.py
```diff
@@ -0,0 +1,169 @@
...
+            'cfg': args2yaml(args),
+            'psw': psw,
+        }
+        if room_id in room.keys():
+            raise ValueError
+        else:
+            room[room_id] = meta_info
+        self._save('room', room)
+
+        # Launch FS
```
rayrayraykk, line 80: I will add `exception_handler` here in the next PR, when launching FS failed.

federatedscope/core/configs/config.py
```diff
@@ -238,37 +239,42 @@ def freeze(self, inform=True):
...
-                        "cfg.wandb.use=True but not install the wandb package")
-                    exit()
-
-        if inform:
+        if save:  # save the final cfg
+            Path(self.outdir).mkdir(parents=True, exist_ok=True)
+            with open(os.path.join(self.outdir, "config.yaml"),
+                      'w') as outfile:
+                from contextlib import redirect_stdout
+                with redirect_stdout(outfile):
```
joneswong, line 247: @yxdyc What's the reason to write files via redirecting the standard output stream?

federatedscope/core/configs/config.py
```diff
@@ -238,37 +239,42 @@ def freeze(self, inform=True):
...
+                        import yaml
+                        cfg_yaml = yaml.safe_load(tmp_cfg.dump())
+                        wandb.config.update(cfg_yaml, allow_val_change=True)
+                    except ImportError:
+                        logger.error(
+                            "cfg.wandb.use=True but not install the wandb "
+                            "package")
+                        exit()
+
+        if inform and save:
```
joneswong, line 264: why not just put this if branch in the above one

federatedscope/core/configs/config.py
yxdyc: No special considerations, just personal convention, and I tested that the way it has been written makes the later `yaml.safe_load()` work correctly.  Perhaps changing the code here into sth. with `yaml.safe_dump()` would make it look more elegant.

joneswong (approved):
LGTM

federatedscope/core/configs/config.py
rayrayraykk: Update accordingly.

PR #265, opened 2022-07-28
[Feature]Prettify organizer via cmd2

rayrayraykk:
1. Use cmd2 as `Client` backend instead of cmd (make logs more pretty).
2. Add interfaces to manage launched tasks.
3. Format help doc string.


joneswong (approved):
LGTM

PR #269, opened 2022-07-29
[Workaround] Fix personalization when module has lazy load hooks

rayrayraykk:
Due to `lazy_load` in the torch module (like `pyg.dense.linear`), some personalization or local method might fail for `load_state_dict` when the key does not exist, this is a workaourd to fix this. We should fix this in `cikm` branch @DavdGao.

And I add some `TODO` for fixing B-dis metric @joneswong , which might also run into bug.

### DIFF
--- a/federatedscope/core/auxiliaries/utils.py
+++ b/federatedscope/core/auxiliaries/utils.py
@@ -500,3 +500,9 @@ def calculate_time_cost(instance_number,
         comm_cost = 0
 
     return comp_cost, comm_cost
+
+
+def merge_param_dict(raw_param, filtered_param):
+    for key in filtered_param.keys():
+        raw_param[key] = filtered_param[key]
+    return raw_param
--- a/federatedscope/core/trainers/torch_trainer.py
+++ b/federatedscope/core/trainers/torch_trainer.py
@@ -17,7 +17,7 @@ from federatedscope.core.trainers.trainer import Trainer
 from federatedscope.core.auxiliaries.dataloader_builder import WrapDataset
 from federatedscope.core.auxiliaries.dataloader_builder import get_dataloader
 from federatedscope.core.auxiliaries.ReIterator import ReIterator
-from federatedscope.core.auxiliaries.utils import param2tensor
+from federatedscope.core.auxiliaries.utils import param2tensor, merge_param_dict
 from federatedscope.core.monitors.monitor import Monitor
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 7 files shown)


federatedscope/core/trainers/torch_trainer.py
```diff
@@ -68,8 +69,10 @@ def update(self, model_parameters, strict=False):
         """
         for key in model_parameters:
             model_parameters[key] = param2tensor(model_parameters[key])
-        self.ctx.model.load_state_dict(self._param_filter(model_parameters),
-                                       strict=strict)
+        # Due to lazy load, we merge two state dict
```
xieyxclack, line 72: When would lazy load be used? And can `merge_param_dict` works well when not using lazy load?

xieyxclack (commented):
LGTM

federatedscope/core/trainers/torch_trainer.py
rayrayraykk: This won't affect when not using lazy load.

federatedscope/core/auxiliaries/utils.py
```diff
@@ -500,3 +500,9 @@ def calculate_time_cost(instance_number,
         comm_cost = 0
 
     return comp_cost, comm_cost
+
+
+def merge_param_dict(raw_param, filtered_param):
+    for key in filtered_param.keys():
+        raw_param[key] = filtered_param[key]
```
DavdGao, line 507: Maybe only cover the attribute `data`?

DavdGao: When the user forgets to filter some certain parameters, let's say `conv.weight`, and the `raw_param` doesn't have the key, maybe it is strange for the user to report KeyError here (Because they don't know why we merge the parameters here). 
So maybe we should raise a detailed error here to explain it?

federatedscope/core/trainers/torch_trainer.py
DavdGao: Maybe explain the reason in more detail. `From torch_geometric version 2.03 ...`

federatedscope/core/worker/server.py
```diff
@@ -447,7 +449,9 @@ def _perform_federated_aggregation(self):
             }
             # logger.info(f'The staleness is {staleness}')
             result = aggregator.aggregate(agg_info)
-            model.load_state_dict(result, strict=False)
+            # Due to lazy load, we merge two state dict
```
DavdGao, line 452: The same as above

DavdGao (commented):
Please see the inline comments

xieyxclack (approved):
Approve

federatedscope/core/auxiliaries/utils.py
rayrayraykk: State dict only contains param, and grads are in model. parameters()

Issue #289, opened 2022-08-06
Bug appears when the elements in a Sampler's attribute self.client_state are all zeros


TheSunWillRise:
I meet a bug when I run a hpo exp by
`python federatedscope/hpo.py --cfg federatedscope/example_configs/femnist/sha_wrap_fedex.yaml`

Traceback (most recent call last):
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/hpo.py", line 43, in <module>
    _ = scheduler.optimize()
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/autotune/algos.py", line 254, in optimize
    current_perfs = self._iteration(current_configs)
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/autotune/algos.py", line 235, in _iteration
    perfs = self._evaluate(configs)
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/autotune/algos.py", line 183, in _evaluate
    perfs[i] = make_trial(trial_cfg)
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/autotune/algos.py", line 34, in make_trial
    results = Fed_runner.run()
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/core/fed_runner.py", line 192, in run
    self._run_simulation_online()
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/core/fed_runner.py", line 228, in _run_simulation_online
    self._handle_msg(msg)
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/core/fed_runner.py", line 420, in _handle_msg
    self.server.msg_handlers[msg.msg_type](msg)
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/autotune/fedex/server.py", line 210, in callback_funcs_model_para
    return self.check_and_move_on()
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/autotune/fedex/server.py", line 363, in check_and_move_on
    self.broadcast_model_para(
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/autotune/fedex/server.py", line 157, in broadcast_model_para
    receiver = self.sampler.sample(size=sample_client_num)
  File "/home/chenganda.cad/codes/FederatedScope/federatedscope/core/sampler.py", line 53, in sample
    sampled_clients = np.random.choice(idle_clients,
[...]

PR #291, opened 2022-08-08
Fix client state error

joneswong:
Some changes have not been reflected in subclasses of server, as #289  reported.


xieyxclack (approved):
LGTM

PR #299, opened 2022-08-08
re-organize aggregators

xieyxclack:
As the title says.


DavdGao (approved):
Looks good to me

federatedscope/core/auxiliaries/aggregator_builder.py
```diff
@@ -10,7 +10,7 @@ def get_aggregator(method, model=None, device=None, online=False, config=None):
         from federatedscope.cross_backends import FedAvgAggregator
         return FedAvgAggregator(model=model, device=device)
     else:
-        from federatedscope.core.aggregator import ClientsAvgAggregator, \
+        from federatedscope.core.aggregators import ClientsAvgAggregator, \
```
DavdGao, line 13: @xieyxclack  It seems like more than one argument will influence the type of aggregator, maybe we should ensure the final type of aggregator and resolve the potential conflicts before Fedrunner.run(), so that only one argument `aggregator.type` is enough.

PR #302, opened 2022-08-08
support using cached data and re-splitting for huggingface datasets

yxdyc:
support 
- using cached data for huggingface datasets;
-  re-splitting the GLUE versions into custom FL versions


federatedscope/core/auxiliaries/data_builder.py
```diff
@@ -285,8 +290,14 @@ def load_torchtext_data(name, splits=None, config=None):
 
         if config.model.type.endswith('transformers'):
             from transformers import AutoTokenizer
-            tokenizer = AutoTokenizer.from_pretrained(
-                config.model.type.split('@')[0])
+
+            try:
```
rayrayraykk, line 294: Why need a try here?

federatedscope/core/auxiliaries/data_builder.py
```diff
@@ -402,6 +413,7 @@ def load_torch_geometric_data(name, splits=None, config=None):
 
     def load_huggingface_datasets_data(name, splits=None, config=None):
         from datasets import load_dataset
+        from datasets import load_from_disk
```
rayrayraykk, line 416: Merge Line 415 and Line 416.

federatedscope/core/auxiliaries/data_builder.py
```diff
@@ -410,18 +422,46 @@ def load_huggingface_datasets_data(name, splits=None, config=None):
...
-        dataset = load_dataset(path=config.data.root,
-                               name=name,
-                               **filtered_args)
+        logger.info("Begin to load huggingface dataset")
+        if "hg_cache_dir" in raw_args:
+            hugging_face_path = raw_args["hg_cache_dir"]
+        else:
+            hugging_face_path = os.getcwd()
+
+        if "load_disk_dir" in raw_args:
```
rayrayraykk, line 431: We could use a try here when `load_disk_dir` in `raw_args` but without the data file cached.

rayrayraykk (commented):
Good job, we could accelerate to load the huggingface data file with the cache.

federatedscope/core/auxiliaries/data_builder.py
yxdyc: In case of the cached file not existed

rayrayraykk (approved):
LGTM

PR #313, opened 2022-08-11
Add scripts for running gcn with dp & fix tensor under different device

rayrayraykk:
As the title says.


joneswong:
@DavdGao hi dawei, we are unfamiliar with this algorithm. Could you give an explanation for us about what is the (\epsilon, \delta) each single query (training round) satisfies, when we specify `constant=1` and `eps` in {50, 500, 5000}? Thanks!

federatedscope/core/trainers/trainer_nbafl.py
```diff
@@ -109,8 +109,9 @@ def inject_noise_in_broadcast(cfg, sample_client_num, model):
 
     # Clip weight
     for p in model.parameters():
-        p.data = p.data / torch.max(torch.ones(size=p.shape),
-                                    torch.abs(p.data) / cfg.nbafl.w_clip)
+        p.data = p.data / torch.max(
+            torch.ones(size=p.shape).to(p.data.device),
```
joneswong, line 113: is it possible to use `torch.ones(device=balabala)` directly?

joneswong (approved):
The shell script looks good.

DavdGao:
> @DavdGao hi dawei, we are unfamiliar with this algorithm. Could you give an explanation for us about what is the (\epsilon, \delta) each single query (training round) satisfies, when we specify `constant=1` and `eps` in {50, 500, 5000}? Thanks!

The $\epsilon-\delta$-DP guarantee promises given two neighbor dataset $D$ and $D'$, $P(M(D)\in S))\leq \exp{\epsilon} P(M(D)\in S) + \delta$, where $M$ is the training process in our setting.

benchmark/FedHPOB/scripts/gcn/cora_dp.yaml
```diff
@@ -0,0 +1,41 @@
+use_gpu: True
+device: 0
+early_stop:
+  patience: 100
+seed: 12345
+federate:
```
DavdGao, line 6: please make sure you don't need to inject noise in download channel. Without `join_in_info`, the server won't inject noise when the clients download the model

benchmark/FedHPOB/scripts/gcn/cora_dp.yaml
```diff
@@ -0,0 +1,41 @@
...
+  type: nodefullbatch_trainer
+eval:
+  freq: 1
+  metrics: ['acc', 'correct', 'f1']
+  split: ['test', 'val', 'train']
+nbafl:
+  use: True
+  mu: 0.0
+  w_clip: 0.1
+  epsilon: 2000
```
DavdGao, line 40: epsilon is so large that the variance of the noise will be too small

DavdGao (commented):

benchmark/FedHPOB/scripts/gcn/cora_dp.yaml
rayrayraykk: > please see the inline comments

Updated accordingly.

rayrayraykk:
> please see the inline comments

Updated accordingly.

federatedscope/core/trainers/trainer_nbafl.py
```diff
@@ -63,8 +63,12 @@ def init_nbafl_ctx(base_trainer):
     ctx.regularizer = get_regularizer(cfg.regularizer.type)
 
     # set noise scale during upload
+    if cfg.trainer.type == 'nodefullbatch_trainer':
+        num_train_data = sum(ctx.train_loader.dataset[0]['train_mask'])
```
DavdGao, line 67: Maybe we should wrap all the dataset/dataloader with a unified attribute `num_samples` in the future.

DavdGao (approved):
LGTM

federatedscope/core/trainers/trainer_nbafl.py
rayrayraykk: Agree.

PR #314, opened 2022-08-11
[Hotfix] Make the message comparator more robust

yxdyc:
fix the message comparator such that the eval results are processed earlier than eval results at later rounds


xieyxclack (approved):
LGTM

PR #317, opened 2022-08-12
[bugfix] Fix one-shot exp utils

rayrayraykk:
As the title says.


federatedscope/autotune/hpbandster.py
```diff
@@ -27,8 +27,8 @@ def get_next_iteration(self, iteration, iteration_kwargs={}):
         # number of configurations in that bracket
         n0 = int(np.floor((self.max_SH_iter) / (s + 1)) * self.eta**s)
         ns = [max(int(n0 * (self.eta**(-i))), 1) for i in range(s + 1)]
-        if os.path.exists(self.working_folder):
-            self.clear_cache()
+        # if os.path.exists(self.working_folder):
```
joneswong, line 30: so how to avoid restoring from the ckpts of previous SHA's attempts?

rayrayraykk: Fixed, but I wonder as we remove ckpts but the `Z` yaml file will also be updated by the next SHA's attempts. How can we avoid this? Maybe we need a discussion.

joneswong (approved):
LGTM

PR #331, opened 2022-08-16
[Scripts] Add scripts for FedProx on Cora

rayrayraykk:


benchmark/FedHPOB/scripts/gcn/cora_prox.yaml
```diff
@@ -0,0 +1,38 @@
...
+    weight_decay: 0.0005
+trainer:
+  type: nodefullbatch_trainer
+eval:
+  freq: 1
+  metrics: ['acc', 'correct', 'f1']
+  split: ['test', 'val', 'train']
+fedprox:
+  use: True
+  mu: 5.0
```
joneswong, line 38: this hyperparameter is critical in this algorithm. I am wondering how we should consider it in our setting, that is to say, to be searched or fixed?

rayrayraykk: `mu` is to be searched in (0.1 1.0 5.0).

joneswong (approved):
approved.

PR #336, opened 2022-08-17
[Feature] Add new system model

rayrayraykk:
1. New system model from:
Wang et al. A field guide to federated optimization. arXiv preprint 460 arXiv:2107.06917, 2021.

2. Refactor some interface.


joneswong (commented):
Please see p35 first two lines

joneswong (approved):
approved.

PR #349, opened 2022-08-24
Update README.md

emirceyani:
Add the SpreadGNN paper and please provide implementation


joneswong:
Thanks for your contribution! There seems to be a conflict, which has to be resolved before we could merge your commit.

PR #369, opened 2022-09-08
[Feature] Add serveral differential privacy algorithms

DavdGao:
Main modifications: 
- Add functional mechanism for differential privacy, including:
    - a trainer named `fmtrainer` 
    - two models for linear regression and logistic regression in funcational mechanism
- Support gaussian and laplace DP mechanism by adding new optimizers
- Add unittest and scripts for DP mechanism

### DIFF
--- a/federatedscope/core/regularizer/l2_regularizer.py
+++ b/federatedscope/core/regularizer/l2_regularizer.py
@@ -1,0 +1,36 @@
+from federatedscope.register import register_regularizer
+from torch.nn import Module
+
+import torch
+
+REGULARIZER_NAME = "l2_regularizer"
+
+
+class L2Regularizer(Module):
+    """Returns the l2 norm of weight
+
+        Arguments:
+            p (int): The order of norm.
+            tensor_before: The original matrix or vector
+            tensor_after: The updated matrix or vector
+
+        Returns:
+            Tensor: the norm of the given udpate.
+    """
+    def __init__(self):
+        super(L2Regularizer, self).__init__()
+
+    def forward(self, ctx):
+        l2_norm = 0.
+        for param in ctx.model.parameters():
+            l2_norm += torch.sum(param**2)
+        return l2_norm
+
+
+def call_l2_regularizer(type):
+    if type == REGULARIZER_NAME:
+        regularizer = L2Regularizer
+        return regularizer
+
+
+register_regularizer(REGULARIZER_NAME, call_l2_regularizer)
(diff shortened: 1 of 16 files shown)


federatedscope/core/regularizer/l2_regularizer.py
```diff
@@ -0,0 +1,36 @@
...
+import torch
+
+REGULARIZER_NAME = "l2_regularizer"
+
+
+class L2Regularizer(Module):
+    """Returns the l2 norm of weight
+
+        Arguments:
+            p (int): The order of norm.
```
xieyxclack, line 13: 'p=2'?

federatedscope/core/regularizer/l2_regularizer.py
```diff
@@ -0,0 +1,36 @@
...
+
+        Returns:
+            Tensor: the norm of the given udpate.
+    """
+    def __init__(self):
+        super(L2Regularizer, self).__init__()
+
+    def forward(self, ctx):
+        l2_norm = 0.
+        for param in ctx.model.parameters():
```
xieyxclack, line 25: An interesting discussion about whether l2 regularization should be applied to beta and gamma of BN layers: https://discuss.pytorch.org/t/weight-decay-in-the-optimizers-is-a-bad-idea-especially-with-batchnorm/16994.
IMO, we can keep the implementation same as that in `torch.norm`

xieyxclack (commented):
LGTM, but I have not checked the implementation details of `Functional Mechanism: Regression Analysis under Differential Privacy`. And please refer to the inline comments for minor suggestions, thx!

federatedscope/core/regularizer/l2_regularizer.py
DavdGao: modified accordingly

federatedscope/core/regularizer/l2_regularizer.py
DavdGao: The current `torch.norm` will calculate the l2 norm for all parameters including bn weight and bias. 
For now, we use the parameter `skip_bn` to aovid calculating l2-norm for bn layer(with 'bn' in its name).

DavdGao: Conflicts are solved

federatedscope/core/regularizer/l2_regularizer.py
taxuexunshang: In the funcational mechanism original paper, the noise addition process is performed only once for the objective function, while your implementation performs noise addition in each iteration, is this supported by the theory?

PR #371, opened 2022-09-08
fix update key not enabled & remove redundant cfg key

rayrayraykk:
## Main changes:
* I make `MetricCalculator` as a sub-module of `Monitor` and fix `update_key` bugs.
* remove redundant code and key of `the larger_the_better` and `the_smaller_the_better`, these would be done in `monitor`.

## Questions:
And I have a question about `ctx` and `monitor`,  as the code block shows:
```python
    def _hook_on_fit_start_calculate_model_size(self, ctx):
        if not isinstance(self.ctx.monitor, Monitor):
            logger.warning(
                f"The trainer {type(self)} does contain a valid monitor, "
                f"this may be caused by initializing trainer subclasses "
                f"without passing a valid monitor instance."
                f"Plz check whether this is you want.")
            return
        if self.ctx.monitor.total_model_size == 0:
            self.ctx.monitor.track_model_size(ctx.models)
```

As `ctx` as args pass to `_hook_on_fit_start_calculate_model_size`, why do we still need use `self.ctx.monitor` rather than `ctx.monitor`? @yxdyc @DavdGao

### DIFF
--- a/federatedscope/contrib/metrics/poison_acc.py
+++ b/federatedscope/contrib/metrics/poison_acc.py
@@ -25,7 +25,8 @@ def load_poison_metrics(ctx, y_true, y_pred, y_prob, **kwargs):
 
 def call_poison_metric(types):
     if 'poison_attack_acc' in types:
-        return 'poison_attack_acc', load_poison_metrics
+        the_larger_the_better = True
+        return 'poison_attack_acc', load_poison_metrics, the_larger_the_better
 
 
 register_metric('poison_attack_acc', call_poison_metric)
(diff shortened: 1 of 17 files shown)


federatedscope/contrib/metrics/poison_acc.py
```diff
@@ -25,7 +25,8 @@ def load_poison_metrics(ctx, y_true, y_pred, y_prob, **kwargs):
 
 def call_poison_metric(types):
     if 'poison_attack_acc' in types:
-        return 'poison_attack_acc', load_poison_metrics
+        the_larger_the_better = True
```
rayrayraykk, line 28: @Osier-Yi Please help me check this metric, thanks!

rayrayraykk:
This PR has been merged to branch `rayrayraykk:refactor_runner`.

PR #374, opened 2022-09-13
Fix early_stop when the metric is the larger the better

rayrayraykk:
As the title says.


yxdyc (commented):
LGTM, we can modify the comment line 65 accordingly, for easy code reading

yxdyc (approved):

PR #379, opened 2022-09-21
[Feature] Support more message type in server

DavdGao:
## Our Target
- We are planning to add a consultation stage before training for differential privacy. 
- We decompose it into several PRs to avoid large modification at one time. 

## What's in this PR
- This PR is to support more message types in the server (currently we only support `train` and `eval` in `self.msgbuffer`)
- We add a enumeration class `MSGBUFFER` in `enums.py` to avoid using string directly
- We modify `check_and_move_on` and `check_buffer` fucntions to support more messages. 
- The modification is compatible with previous versions

### DIFF
--- a/federatedscope/core/auxiliaries/enums.py
+++ b/federatedscope/core/auxiliaries/enums.py
@@ -11,6 +11,12 @@ class MODE:
     FINETUNE = 'finetune'
 
 
+class MSGBUFFER:
+    TRAIN = 'train'
+    EVAL = 'eval'
+    CONSULT = 'consult'
+
+
 class TRIGGER:
     ON_FIT_START = 'on_fit_start'
     ON_EPOCH_START = 'on_epoch_start'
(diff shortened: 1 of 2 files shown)


federatedscope/core/workers/client.py
```diff
@@ -225,12 +226,12 @@ def callback_funcs_for_model_para(self, message: Message):
             # A fragment of the shared secret
             state, content, timestamp = message.state, message.content, \
                                         message.timestamp
-            self.msg_buffer['train'][state].append(content)
+            self.msg_buffer[STAGE.TRAIN][state].append(content)
```
rayrayraykk, line 229: I'm not sure whether the 'stage' is an appropriate key name, as the key of `msg_buffer` should be message type (although it is not now: `msg_buffer`['train'] contains model, `msg_buffer`['eval'] contains eval results).

Maybe we should use the true message type as the key of `msg_buffer`.

rayrayraykk (commented):
The other part looks good to me.

PR #381, opened 2022-09-22
Add HPOBench as backend demo for FedHPOBench

rayrayraykk:
# Main changes
* HPOBench is a very popular and influential library in the HPO field, so we provide serval demos for exposing FedHPOBench through the interfaces of HPOBench in `demo\`. Add Tabular and Surrogate FEMNIST demo with HPOBench backend.
* Add interface of fairness-related regular terms in the tabular benchmark.


joneswong (approved):
LGTM

PR #365, opened 2022-09-06
Refactor data-related interfaces & add interfaces for trainer and worker

rayrayraykk:
# Main Changes

# Data

## Data module overview

`federatedscope.core.auxiliaries.data_builder.get_data` **-700 lines!**

* **Load Dataset** (`federatedscope.core.data.utils.load_dataset`): 
  * Load local file to torch dataset, **-300 lines!**
* **Translate data** (`federatedscope.core.data.BaseDataTranslator`)
  * Dataset -> ML split -> FL split -> DataLoader for FS
* **Convert mode**(`federatedscope.core.data.utils.convert_data_mode`)
  * To adapt simulation mode and distributed mode

## Data Translator

Dataset -> (ML split) -> (FL split) -> DataLoader for FS

* **ML split**(`split_train_val_test`): 
  * Build train/val/test
* **FL split**(`split_to_client`): 
  * Build Data for each client

## Data interface

* **ClientData**(`federatedscope.core.data.ClientData`)

  * A subclass of `dict` with `train`, `val` and `test`.

  * Convert dataset to DataLoader.

    * `cfg.dataloader.type`, `cfg.dataloader.batch_size`, `cfg.dataloader.shuffle`

  * Example:

    * ```python
      # Instantiate client_data for each Client
      client_data = ClientData(PyGDataLoader, 
                               cfg, 
                               train=train_data, 
                               val=None, 
                               test=test_data)
      # other_cfg with different batch size
      client_data.setup(other_cfg)
      print(client_data)
      
      >> {'train': PyGDataLoader(train_data), 'test': PyGDataLoader(test_data)}
      ```

* **StandaloneDataDict**(`federatedscope.core.data.StandaloneDataDict`)

  * A subclass of `dict` with **client_id** as keys: 
    * `{1: ClientData, 2: ClientData, ...}`
  * Responsible for training/evaluation method conversion：
    * Global evaluation, global training, etc.

# Trainer

* BaseTrainer

  * Code without hook-like functions (For starters)

  * Example:

    ```python
    class BaseTrainer(abc.ABC):
        def __init__(self, model, data, device, **kwargs):
            self.model = model
            self.data = data
            self.device = device
            self.kwargs = kwargs
    
        @abc.abstractmethod
        def train(self):
            raise NotImplementedError
    
        @abc.abstractmethod
        def evaluate(self, target_data_split_name='test'):
            raise NotImplementedError
    
        @abc.abstractmethod
        def update(self, model_parameters, strict=False):
[...]

PR #388, opened 2022-09-29
Add checks for completeness of msg_handler

rayrayraykk:
# Main changes
Add a check for completeness of msg_handler in workers. (Required `networkx` and `matplotlib` and I add them in `dev` mode)
## What's New in worker
A `classmethod`: `get_msg_handler_dict`;
A `self.msg_handlers_str` to store triplets <`msg_in`, `handler_name`, `[msg_out0, msg_out1, ...]`>

## How to use:
Set `cfg.check_completeness=True`:
```
python federatedscope/main.py --cfg scripts/example_configs/femnist.yaml check_completeness True
```

## What information will the user know:
If the checks are enabled, three types of info will show: `pass`, `warning`, `error` for the case:
completeness

And a DiGraph will be plotted in `exp_dir` to remind users something might goes wrong:
image

image

For pure FedAvg:
image


rayrayraykk:
@xieyxclack Please help me check the `out_msg` type in `_register_default_handlers`, thanks!

federatedscope/core/fed_runner.py
```diff
@@ -427,3 +432,62 @@ def _handle_msg(self, msg, rcv=-1):
...
                     download_bytes)
+
+    def check(self):
+        """
+        Check the completeness of Server and Client.
+
+        Returns:
+
+        """
+        if self.cfg.check_completeness:
```
joneswong, line 443: I don't think it is necessary to indent such a huge block of code

federatedscope/core/fed_runner.py
```diff
@@ -427,3 +432,62 @@ def _handle_msg(self, msg, rcv=-1):
...
+                            f'Completeness check raises warning for '
+                            f'some handlers not in FL process! Save '
+                            f'check results in {fig_path}.')
+                else:
+                    logger.error(f'Completeness check fails for there is no'
+                                 f'path from `join_in` to `finish`! Save '
+                                 f'check results in {fig_path}.')
+            except Exception as error:
+                logger.warning(f'Completeness check failed for {error}!')
+        return
```
joneswong, line 493: so the reason is that we cannot say yes or not about the correctness/completeness, right?

rayrayraykk: As the document shows, the correctness/completeness checks only check the message-handler pairs and raise three types of logs (info, warning, and error). If something goes wrong with Python code, we'd better keep the exception stack as it is. So the return value is meaningless.

federatedscope/core/fed_runner.py
rayrayraykk: Update accordingly.

xieyxclack (approved):
LGTM

rayrayraykk:
DO NOT merge this PR ahead of #365 .

xieyxclack (approved):

PR #408, opened 2022-10-24
[Feature] Refactor FedRunner and optimize trainer module

rayrayraykk:
## What's changed
1. Runner -> StandaloneRunner + DistributedRunner
2. Add README for Trainer module
3. Make `num_val_batch`, `num_total_train_batch`.... (related to `cfg`) as @property method (Users could still use `setattr` or `=` to change these values, but the changes will make the `@property` function invalid.)
```python
@property
def num_total_train_batch(self):
    if self.get('num_total_train_batch'):
        return self.get('num_total_train_batch')
    return self._calculate_batch_epoch_num(mode='train')[3]
```
5. Optimize doc string for hook functions and API ref.
6. Bind each metric with one and only key `THE_LARGER_THE_BETTER`
7. Move functions in `utils` to `xxx.utils`

## Details
Coming soon.


rayrayraykk:
Dicard this PR for the changes containing in https://github.com/alibaba/FederatedScope/pull/415 .

PR #415, opened 2022-10-28
[Feature] Refactor FedRunner, optimize trainer module and optimize CI

rayrayraykk:
## What's changed
1. Runner -> StandaloneRunner + DistributedRunner
2. Add README for Trainer module
3. Make `num_val_batch`, `num_total_train_batch`.... (related to `cfg`) as @property method (Users could still use `setattr` or `=` to change these values, but the changes will make the `@property` function invalid.)
```python
@property
def num_total_train_batch(self):
    if self.get('num_total_train_batch'):
        return self.get('num_total_train_batch')
    return self._calculate_batch_epoch_num(mode='train')[3]
```
5. Optimize doc string for hook functions and API ref.
6. Bind each metric with one and only key `THE_LARGER_THE_BETTER`
7. Move functions in `utils` to `xxx.utils`
8. Optimize CI, as you can see below.

## Details
Coming soon.


federatedscope/core/auxiliaries/metric_builder.py
```diff
@@ -16,6 +16,6 @@ def get_metric(types):
     for func in register.metric_dict.values():
         res = func(types)
         if res is not None:
-            name, metric = res
```
joneswong, line 19: I just noticed that `get_metric` has no docstring and returns a dict that has such a special design of its values. Could you just supplement the missing docstring?

federatedscope/core/auxiliaries/trainer_builder.py
```diff
@@ -108,6 +108,11 @@ def get_trainer(model=None,
             raise ValueError('Trainer {} is not provided'.format(
                 config.trainer.type))
 
+    if not isinstance(trainer, Trainer):
```
joneswong, line 111: It might be hard for users to interpret this warning message. At least I need a second-thought

federatedscope/core/fed_runner.py
```diff
@@ -14,6 +15,457 @@
 logger = logging.getLogger(__name__)
 
 
+def get_runner(data, server_class, client_class, config, client_configs=None):
+    # Instantiate a Runner based on a configuration file
+    mode = config.federate.mode.lower()
+    runner_dict = {
```
joneswong, line 21: why creating a dict just for branching?

joneswong (commented):
good job! as the changes are huge, is it possible to supplement UT cases in this pr?

scripts/distributed_scripts/distributed_configs/distributed_client_1.yaml
```diff
@@ -12,7 +12,7 @@ distribute:
   client_host: '127.0.0.1'
   client_port: 50052
   role: 'client'
-  data_file: 'toy_data/client_1_data'
+  data_idx: 1
```
xieyxclack, line 15: TODO: provide an example that loads data from separate data files in distributed mode

xieyxclack (commented):
LGTM, and thanks a lot for such great work on the refactoring and documenting! @rayrayraykk

joneswong (approved):
Approved.

federatedscope/core/trainers/context.py
```diff
@@ -117,51 +152,107 @@ def __init__(self,
...
-            # Calculate the number of update steps during training given the
-            # local_update_steps
-            self.num_train_batch, self.num_train_batch_last_epoch, \
-                self.num_train_epoch, self.num_total_train_batch = \
+    # Train related property, query from `cfg` if not set
+    @property
+    def num_train_batch(self):
+        if self.get('num_train_batch'):
+            return self.get('num_train_batch')
+        return self._calculate_batch_epoch_num(mode='train')[0]
```
DavdGao, line 176: The number of epoch is decided by the selected dataset. Maybe we should name the parameter as `split` or `dataset` here rather than `mode`

federatedscope/core/trainers/context.py
```diff
@@ -117,51 +152,107 @@ def __init__(self,
...
+
+    @property
+    def num_train_batch_last_epoch(self):
+        if self.get('num_train_batch_last_epoch'):
+            return self.get('num_train_batch_last_epoch')
+        return self._calculate_batch_epoch_num(mode='train')[1]
```
DavdGao, line 182: the same as above

federatedscope/core/trainers/context.py
```diff
@@ -46,7 +45,9 @@ def clear(self, lifecycle):
 
 
 class Context(LifecycleDict):
-    """Record and pass variables among different hook functions
+    """
```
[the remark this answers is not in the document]
DavdGao: Should we support any combination of different `mode`(train/finetune/eval) and different `split` (dataset)? Or customized name of split, e.g. `"train1"`

DavdGao (commented):
Please see the inline comments.

DavdGao:
BTW, maybe we should unify the name of directories, e.g. `trainers` or `trainer`/`workers` or `worker`/`criterion` or `loss` in different tasks(core/cv/gfl/nlp/mf)

PR #422, opened 2022-11-04
add one new paper

qbc2016:
as the title says


materials/paper_list/FL-Tree/README.md
```diff
@@ -1,27 +1,29 @@
-# Federated Learning for Tree
+## Federated Learning for Tree
```
xieyxclack, line 1: Maybe we should change it to "Federated Learning for Tree-based Models"

qbc2016: fine

xieyxclack (approved):

PR #426, opened 2022-11-10
Add condition param to autotune; Add feature engineering module

rayrayraykk:
# Main changes
## condition param
The style are shown below:
```bash
   xgb_base.use  train.optimizer.lr  train.optimizer.num_of_trees  performance
0          True                 NaN                           5.0     0.59876
1         False            0.694859                           NaN     0.599826
2         False            0.876475                           NaN     0.600004

```

## wandb
How to enable:
Install wandb and setup init configs for wandb
Run with `wandb.use True`:
``` bash
python federatedscope/hpo.py --cfg federatedscope/autotune/baseline/fedex_vfl.yaml wandb.use True wandb.name_project vfl_demo wandb.name_user vfl_demo
```

## Feature engineering
I implement Feature engineering as a wrapper for workers.

```
FederatedScope
├── federatedscope
│   ├── core           
│   |   ├── ...
│   |   ├── feature
│   |   |   ├── hFL
│   |   |   |   ├── ...
│   |   |   ├── vFL
│   |   |   |   ├── preprocess
│   |   |   |   |   ├── log_transform.py
│   |   |   |   |   ├── quantile_binning.py
│   |   |   |   |   ├── ...
│   |   |   |   ├── selection
│   |   |   |   |   ├── correlation_filter.py
│   |   |   |   |   ├── iv_filter.py
│   |   |   |   |   ├── ...
```

How to use:
```bash
python federatedscope/main.py --cfg federatedscope/vertical_fl/xgb_base/baseline/xgb_base_on_adult.yaml feat_engr.type vfl_instance_norm
```


federatedscope/autotune/algos.py
```diff
@@ -57,7 +57,7 @@ def run(self):
                                server_class=get_server_cls(self._trial_cfg),
                                client_class=get_client_cls(self._trial_cfg),
                                config=self._trial_cfg.clone(),
-                               client_config=self._client_cfgs)
+                               client_configs=self._client_cfgs)
```
joneswong, line 60: to avoid such kinds of mistakes, we need to improve the coverage of our unit test cases

joneswong (approved):
good job! conditional search space is prevalent, and we really need it.

federatedscope/autotune/algos.py
```diff
@@ -29,10 +28,10 @@ def make_trial(trial_cfg, client_cfgs=None):
                             server_class=get_server_cls(trial_cfg),
                             client_class=get_client_cls(trial_cfg),
                             config=trial_cfg.clone(),
-                            client_config=client_cfgs)
+                            client_configs=client_cfgs)
```
joneswong, line 31: we'd better provide UT cases to cover this module.

federatedscope/autotune/algos.py
```diff
@@ -29,10 +28,10 @@ def make_trial(trial_cfg, client_cfgs=None):
...
     results = Fed_runner.run()
     key1, key2 = trial_cfg.hpo.metric.split('.')
-    return results[key1][key2]
+    return results[key1][key2], results
```
joneswong, line 34: please supplement docstring for this function since we should let users know these changes in returned value(s).

joneswong: BTW, this makes the first returned value redundant...

federatedscope/autotune/baseline/fedex_vfl.yaml
```diff
@@ -0,0 +1,53 @@
+use_gpu: False
```
joneswong, line 1: FedEx is designed for horizontal scenarios, imo. When clients have totally different inputs, I am wondering whether updating a policy based on all their feedback is still a reasonable trick. I suggest you (and @qbc2016 ?) to have a glance at <FLoRA>.

federatedscope/autotune/utils.py
```diff
@@ -171,7 +255,32 @@ def eval_in_fs(cfg, config, budget, client_cfgs=None):
...
+    try:
+        import wandb
+        key1, key2 = trial_cfg.hpo.metric.split('.')
+        log_res = {
+            'Trial_index': trial,
+            'Config': config_bool2int(config),
+            trial_cfg.hpo.metric: results[key1][key2],
+        }
+        wandb.log(log_res)
+    except ImportError:
```
joneswong, line 284: I didn't see any necessity to have a try-catch block from this exception handling step.

federatedscope/core/data/base_data.py
```diff
@@ -149,9 +149,9 @@ class ClientData(dict):
     """
     def __init__(self, client_cfg, train=None, val=None, test=None, **kwargs):
         self.client_cfg = None
-        self.train = train
-        self.val = val
-        self.test = test
+        self.train_data = train
```
joneswong, line 152: I don't think these changes are necessary as the names of the corresponding arguments of this constructor have not been changed. It would be better to provide a docstring for this constructor to elaborate on the arguments.

federatedscope/hpo.py
```diff
@@ -26,6 +26,11 @@
     cfg_opt, client_cfg_opt = parse_client_cfg(args.opts)
     init_cfg.merge_from_list(cfg_opt)
 
+    # Update Exp_name for hpo
+    if init_cfg.expname == '':
+        from federatedscope.autotune.utils import generate_hpo_exp_name
```
joneswong, line 31: does this matter anda's exp?

joneswong (commented):
generally, looks good. please resolve these minor concerns. thanks!

federatedscope/core/configs/cfg_data.py
```diff
@@ -64,6 +64,23 @@ def extend_data_cfg(cfg):
     cfg.data.quadratic.min_curv = 0.02
     cfg.data.quadratic.max_curv = 12.5
 
+    # feature engineer
```
xieyxclack, line 67: It should be ''feature engineering''?

federatedscope/core/configs/cfg_fl_setting.py
```diff
@@ -67,10 +67,10 @@ def extend_fl_setting_cfg(cfg):
     # ---------------------------------------------------------------------- #
     # Vertical FL related options (for demo)
     # ---------------------------------------------------------------------- #
+    cfg.vertical_dims = [5, 10]  # Avoid to be removed when `use` is False
```
xieyxclack, line 70: Why we still need `vertical_dims` if `vertical.use=False`?

federatedscope/core/feature/vfl/preprocess/instance_norm.py
```diff
@@ -0,0 +1,305 @@
...
+            self.comm_manager.send(
+                Message(msg_type='instance_var',
+                        sender=self.ID,
+                        receiver=list(self.comm_manager.neighbors.keys()),
+                        timestamp=self.cur_timestamp,
+                        content=instance_var))
+            self._init_data_related_var()
+            self.trigger_train_func(**self.kwargs_for_trigger_train_func)
+
+    # Bind method to instance
```
xieyxclack, line 110: It looks good! @qbc2016 Maybe we can improve some modules in FL-Tree like this

xieyxclack (commented):
The feature engineering module looks good to me!

federatedscope/autotune/run.py
```diff
@@ -0,0 +1,14 @@
+def run_scheduler(scheduler, cfg, client_cfgs=None):
```
joneswong, line 1: please provide docstring

federatedscope/autotune/utils.py
```diff
@@ -166,11 +256,32 @@ def eval_in_fs(cfg, config, budget, client_cfgs=None):
...
+    import copy
+    new_dict = copy.deepcopy(config)
+    for key, value in new_dict.items():
+        if isinstance(new_dict[key], bool):
+            new_dict[key] = int(value)
+    return new_dict
+
+
+def log2wandb(trial, config, results, trial_cfg):
+    import wandb
```
joneswong, line 280: As I have no idea about the overhead of repeatedly importing a module, I am wondering is it appropriate to do this? Have you benchmarked such a behavior? Here is a related thread: https://stackoverflow.com/questions/128478/should-import-statements-always-be-at-the-top-of-a-module

federatedscope/core/configs/cfg_fl_setting.py
joneswong: so what is this? I guess it is the input dim. If it is the case, why do we need it?

federatedscope/core/fed_runner.py
```diff
@@ -160,7 +164,7 @@ def _setup_server(self, resource_info=None, client_resource_info=None):
                 wrap_nbafl_server
             wrap_nbafl_server(server)
         logger.info('Server has been set up ... ')
-        return server
+        return self.feat_engr_wrapper_server(server)
```
joneswong, line 167: It is ok to inject those feature engineering procedures into an FL course in this way, but we shall change the instantiation of workers to a better (more general and unified) way. Actually, it is quite confusing to wrap a worker by feature engineering-related wrapper. Feature engineering is just a tiny step in an FL course, which doesn't change a worker significantly. One usual way to instantiate a worker from a collection of such pluggable behaviors is to use factory pattern I guess.

joneswong (dismissed):
approved.

joneswong (approved):
LGTM

PR #429, opened 2022-11-15
[Feature] krum aggregation rule

xieyxclack:
As the title says.


joneswong (approved):
approved. it seems that there is a `deepcopy` in the aggregation step. [TODO] check whether krum aggregator is consistent with the shared mode.

joneswong:
definitely, please pass UT

PR #435, opened 2022-11-16
Add xgb datasets

qbc2016:
add three new datasets for xgb algo: 'abalone', 'givemesomecredit', 'blogfeedback'


xieyxclack:
This pr can be closed since the modifications have been included in #439

PR #437, opened 2022-11-17
Dlg yaml file change

Osier-Yi:
1. The original version of dlg cannot successfully reconstruct the training data because of the default dropout =0.5 in the convnet2. In the modified yaml file: scripts/attack_exp_scripts/privacy_attack/reconstruct_fedavg_opt_on_femnist.yaml, we set model.dropout=0.
2.  in the unitest of dlg, we also check whether the dlg reconstruction loss is less than the threshold.


xieyxclack (approved):
LGTM

PR #444, opened 2022-11-28
set pin_memory to False to avoid OOM

rayrayraykk:
`pin_memory=True`

WARNING

Using [automatic memory pinning](https://pytorch.org/docs/stable/data.html#memory-pinning) (i.e., setting pin_memory=True), which enables fast data transfer to CUDA-enabled GPUs.


joneswong (approved):
approved.

PR #453, opened 2022-12-05
[WIP] add data, model, and algo that FedSAM needs

joneswong:
We are reproducing the experiments of FedSAM.


joneswong:
BTW, we fix a bug in calling registered splitters.

PR #461, opened 2022-12-06
Fix several bugs

rayrayraykk:
1. Fix conflicts comments in the early stop
2. Allow specifying respective data transformations for different splits in torchvision
3. Avoid using `list` of datasets which will cause the transform functions to fail.


joneswong (dismissed):
LGTM. We shall make the train/val/test split translation more general in the future.

joneswong (approved):
looks better! Now the "merge data" procedure is clear.

PR #463, opened 2022-12-07
fix b-local dissim

rayrayraykk:
# Changes

1. fix b-local dissim
2. Add L2 dissim
3. Fix client-wise split not enabled

Set `eval.monitoring=['blocal_dissim', 'l2_dissim']` to enable this feature


joneswong (approved):
Approved. I strongly suggest to keep each pr focusing on one purpose.

PR #466, opened 2022-12-08
Enable empty split

rayrayraykk:


joneswong:
> In my opinion, when there is no train split, users should be responsible for failure of accessing train split. So this behavior is ok:
> 
> https://github.com/alibaba/FederatedScope/blob/e2ce0cec41bdc320d1e0ae82988cdbef8fef1b48/federatedscope/contrib/trainer/torch_example.py#L41
> 
> As for val and test splits, what is our intended behavior?

e.g., is this still ok? https://github.com/alibaba/FederatedScope/blob/e2ce0cec41bdc320d1e0ae82988cdbef8fef1b48/federatedscope/contrib/trainer/torch_example.py#L68

joneswong (dismissed):
LGTM

joneswong (approved):
approved. It seems that we'd better make our proj more "typed".

PR #475, opened 2022-12-16
move "batch_size" from data to dataloader

qbc2016:
move the key "batch_size" to inside the key "dataloader" in yaml files for xgb, and  fix the related bugs


xieyxclack (approved):
LGTM

PR #490, opened 2023-01-12
modify the import source of the update_logger function

private-mechanism:
There are two modifications. One is updating the import source of the update_logger function, the other is modifying the definition of call_my_data function from "call_my_data(config)" to "call_my_data(config, client_cfgs=None)".


xieyxclack:
This pr is closed since it is same as #491

PR #503, opened 2023-01-31
Add defense algo and sampler

xieyxclack:
The modifications included in this pr:
- Add a new sampler ResponsivenessRealtedSampler
- Add defense algo Krum:
   - To test the effectiveness of Krum, I add a gaussian attacker and change the expected types of `cfg.attack.attacker_id` from `int` to `int/list`, please double check this @Osier-Yi 
   - And to fix the issue caused by merging `int` and `list` configuration items, I add a forced replacement rule to yacs, please double check this @yxdyc 

BTM, these features have been developed and pushed into the branch submit_dev_branch before. This pr focuses on how to add a unit test for ensuring their effectiveness and promotes further development.


joneswong (approved):
LGTM

federatedscope/core/configs/yacs_config.py
```diff
@@ -574,6 +574,14 @@ def conditional_cast(from_type, to_type):
         else:
             return False, None
 
+    # int <-> list, forced replacement
+    # To allow single <-> multiple cases
+    # e.g., attack_id = 1  <->  attack_id = [1,2,3]
+    if replacement_type == int and original_type == list:
+        return replacement
+    if replacement_type == list and original_type == int:
+        return replacement_type
```
yxdyc, line 583: return replacement

federatedscope/core/configs/yacs_config.py
```diff
@@ -574,6 +574,14 @@ def conditional_cast(from_type, to_type):
         else:
             return False, None
 
+    # int <-> list, forced replacement
+    # To allow single <-> multiple cases
+    # e.g., attack_id = 1  <->  attack_id = [1,2,3]
+    if replacement_type == int and original_type == list:
+        return replacement
+    if replacement_type == list and original_type == int:
```
yxdyc, line 582: It is reasonable to convert int into list, such as `1 -> [1]`. However, for the reverse conversion, we may limit it in the case where the list contains only one element?  i.e., `if replacement_type == list and original_type == int and len(replacement) == 1:`

xieyxclack: Here is an example of how this forced replacement works: 
- By default, we have `cfg.attack.attacker_id = -1` to denote that there exists no attacker in the simulated FL course. We can easily change it to `cfg.attack.attacker_id = 1` to conduct experiments with one attacker client#1
- However, when we want multiple attackers, such as setting `cfg.attack.attacker_id = [1,2,3]`, an error would be raised by yacs since the default value (`-1`, type `int`) could be replaced with a new value (`[1,2,3]`, type `list`).   This error motivates me to add such a forced replacement rule to support both single and multiple attackers here.
- Thus, if we only support `1 -> [1]` and `[1] -> 1`, but not for `-1 -> [1,2,3]`, this forced replacement (which is used in the auto-merging and but not an equivalent conversion) is meaningless, IMO.

AnonyMLResearcher (approved):
LGTM

PR #507, opened 2023-02-02
add several recent works on untargeted attacks in FL

private-mechanism:
add several recent works on untargeted attacks in FL


xieyxclack (approved):
good job!

PR #510, opened 2023-02-03
Add gbdt

qbc2016:
I modified the GBDT algo, here are three comments:
(1) the only two differences with XGB in class Tree are gain computation and weight computation
(2) the tree model in GBDT dose not need hessian, but the Trainer class contains the hessian computation, for convenience, I did not modify Trainer, so I still keep the hessian in class Tree
(3) as (2) says, in class Tree_GBDT, in order to not modify the Trainer class, I copied function update_child from Tree_XGB, though in fact, Tree_GBDT does not need to compute hess


xieyxclack (approved):
LGTM

PR #515, opened 2023-02-08
Fix bug in transfomer_builder

rayrayraykk:
We got a bug in transfomer_builder before when all splits transformer enabled, the builder will return {}, which is inconsisdent with the comment. I fix it in this PR to make it consisdent.


joneswong (approved):
approved.

PR #516, opened 2023-02-09
modify the output channel of convnet2 model for cifar10 dataset

private-mechanism:
modify the output channel of convnet2 model for cifar10 dataset


xieyxclack:
This pr is closed since these modifications have been included in #513 , thanks!

Issue #517, opened 2023-02-10
Seeking Flat Minima 


joneswong:
Stochastic Weight Averaging (SWA) has been widely regarded as an effective way to find a solution lying a wide valley. I guess It is convenient to add this feature in FS.
Here is the pseudo code from <Improving Generalization in Federated Learning by Seeking Flat Minima>, which adopt SWA in FL:
![CA6AC50C-7FA0-4CE7-AFDC-85E395A9126D](https://user-images.githubusercontent.com/6416895/217984467-624db9d1-bf99-4e6a-9e82-2894037068e8.png)

PR #519, opened 2023-02-10
Add SWA

rayrayraykk:
This PR is linked to #517 .

Besides SWA, I support `merge_val` in global evaluation.

### DIFF
--- a/federatedscope/core/configs/cfg_fl_algo.py
+++ b/federatedscope/core/configs/cfg_fl_algo.py
@@ -26,7 +26,7 @@ def extend_fl_algo_cfg(cfg):
     cfg.fedprox.mu = 0.
 
     # ---------------------------------------------------------------------- #
-    # fedswa related options, a general fl algorithm
+    # fedswa related options, Stochastic Weight Averaging (SWA)
     # ---------------------------------------------------------------------- #
     cfg.fedswa = CN()
     cfg.fedswa.use = False
--- a/federatedscope/core/configs/cfg_fl_algo.py
+++ b/federatedscope/core/configs/cfg_fl_algo.py
@@ -26,6 +26,14 @@ def extend_fl_algo_cfg(cfg):
     cfg.fedprox.mu = 0.
 
     # ---------------------------------------------------------------------- #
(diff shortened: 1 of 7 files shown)


federatedscope/core/configs/cfg_fl_algo.py
```diff
@@ -25,6 +25,14 @@ def extend_fl_algo_cfg(cfg):
     cfg.fedprox.use = False
     cfg.fedprox.mu = 0.
 
+    # ---------------------------------------------------------------------- #
```
joneswong, line 28: I am wondering whether it is appropriate to put SWA here. It is not an FL algorithm but just a trick to produce another solution from those in the late stage of a training course.

federatedscope/core/fed_runner.py
```diff
@@ -166,6 +166,9 @@ def _setup_server(self, resource_info=None, client_resource_info=None):
         if self.cfg.vertical.use:
             from federatedscope.vertical_fl.utils import wrap_vertical_server
             server = wrap_vertical_server(server, self.cfg)
+        if self.cfg.fedswa.use:
```
joneswong, line 169: This implementation implies that, in the future, we'd better refactor the workers so that we would not need to make another worker class just for adding such a trick.

federatedscope/core/fed_runner.py
```diff
@@ -869,7 +872,7 @@ def _setup_server(self, resource_info=None, client_resource_info=None):
 
         logger.info('Server has been set up ... ')
 
-        return server
+        return self.feat_engr_wrapper_server(server)
```
joneswong, line 875: what's the purpose?

federatedscope/core/fed_runner.py
```diff
@@ -926,7 +929,7 @@ def _setup_client(self,
         else:
             logger.info(f'Client {client_id} has been set up ... ')
 
-        return client
+        return self.feat_engr_wrapper_client(client)
```
joneswong, line 932: what's the purpose and what's the relation to SWA?

joneswong (approved):
approved.

PR #522, opened 2023-02-13
Bind lager_better with each metric

rayrayraykk:
This PR is part of #486 .


joneswong (approved):
Approved. Some functionalities are required now. The `trial_id` arg is not a good design imo.

PR #499, opened 2023-01-18
[WIP]Label-based mode for Xgb

qbc2016:

PR #528, opened 2023-02-20
label-based Tree model for vfl

xieyxclack:
This pr is adapted from #499 

Done:
- Label-based XGB without protection
- Enable protection algos (he for protecting grad_and_hess)

Todo:
- Modules as plugins
- Docs


qbc2016 (approved):
Good job!

PR #530, opened 2023-02-23
[bugfix] Add funcs to check empty subtree and remove the interaction between split position and activate indicator

xieyxclack:
As the title says.


qbc2016 (approved):
LGTM.

PR #533, opened 2023-02-27
Fix bucket size

xieyxclack:
As the title says. 
Note that if one of the bucket is empty, the func `np.concatenate` might return float value array when concatenating [IntList, EmptyList, ...,]


qbc2016 (approved):
LGTM.

PR #536, opened 2023-03-06
Fix python version in dockerfile

rayrayraykk:
As the title says, the python version in [latest] Conda is 3.10, which is not usable for FS.


yxdyc (approved):
LGTM

PR #537, opened 2023-03-07
Bugfix for gbdt and acceleration for op boost algo 

qbc2016:
-) fix gain computation function bug for gbdt
-) add label-based model for gbdt
-) add bucket algo to accelerate for op_boost method


xieyxclack (approved):
LGTM

PR #544, opened 2023-03-15
Optimize MF model

rayrayraykk:
~~Optimize MF model forward:~~

* ~~Before: (user_embedding * item_embedding) [mask]~~
* ~~Now: (user_embedding[user_mask].to_sparse() * item_embedding[item_mask].to_sparse()[mask]~~

Optimization of MF

1. Convert MF model to `nn.Embedding` Version: forward with used embeddings
2. Saving client data to `processed_data.pkl` to save time when repeating the exp.

### DIFF
--- a/federatedscope/mf/dataset/movielens.py
+++ b/federatedscope/mf/dataset/movielens.py
@@ -35,8 +35,8 @@ class VMFDataset:
             test_ratings_all.append(test_ratings)
         # Server holds all
         data[0] = {
-            "train": sp.vstack(train_ratings_all).tocsc(),
-            "test": sp.vstack(test_ratings_all).tocsc()
+            "train": sp.hstack(train_ratings_all).tocsc(),
+            "test": sp.hstack(test_ratings_all).tocsc()
         }
         with open(self.processed_data, 'wb') as f:
             pickle.dump(data, f)
--- a/federatedscope/mf/dataset/movielens.py
+++ b/federatedscope/mf/dataset/movielens.py
@@ -5,6 +5,7 @@ import logging
 from torchvision.datasets.utils import check_integrity, \
     download_and_extract_archive, calculate_md5
 import pandas as pd
+from tqdm import tqdm
 from numpy.random import shuffle
 from scipy.sparse import coo_matrix
 from scipy.sparse import csc_matrix
@@ -23,12 +24,14 @@ class VMFDataset:
         shuffle(id_item)
         items_per_client = np.array_split(id_item, num_client)
         data = dict()
-        for clientId, items in enumerate(items_per_client):
+        for clientId, items in tqdm(enumerate(items_per_client)):
             client_ratings = ratings[:, items]
             train_ratings, test_ratings = self._split_train_test_ratings(
                 client_ratings, test_portion)
             data[clientId + 1] = {"train": train_ratings, "test": test_ratings}
-        self.data = data
+        with open(self.processed_data, 'wb') as f:
+            pickle.dump(data, f)
+        return data
 
 
 class HMFDataset:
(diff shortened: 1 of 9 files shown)


federatedscope/mf/model/model.py
```diff
@@ -15,47 +14,34 @@ class BasicMFNet(Module):
...
-        self.register_parameter('embed_user', self.embed_user)
-        self.embed_item = Parameter(
-            torch.normal(mean=0,
-                         std=0.1,
-                         size=(num_item, num_hidden),
-                         requires_grad=True,
-                         dtype=torch.float32))
-        self.register_parameter('embed_item', self.embed_item)
+        self.num_user, self.num_item = num_user, num_item
+        self.embed_user = torch.nn.Embedding(num_user, num_hidden, sparse=True)
```
joneswong, line 18: I am wondering why sparse is needed?

federatedscope/mf/model/model.py
```diff
@@ -15,47 +14,34 @@ class BasicMFNet(Module):
...
-                                        device=pred.device,
-                                        dtype=torch.float32).to_dense()
-        mask = torch.sparse_coo_tensor(indices,
-                                       np.ones(len(ratings)),
-                                       size=pred.shape,
-                                       device=pred.device,
-                                       dtype=torch.float32).to_dense()
-
-        return mask * pred, label, float(np.prod(pred.size())) / len(ratings)
+        device = self.embed_user.weight.device
```
joneswong, line 22: what indices and ratings are should be stated

federatedscope/mf/model/model.py
```diff
@@ -15,47 +14,34 @@ class BasicMFNet(Module):
...
+        indices = torch.tensor(np.array(indices)).to(device)
+
+        user_embedding = self.embed_user(indices[0])
+        item_embedding = self.embed_item(indices[1])
+
+        pred = torch.diag(torch.matmul(user_embedding, item_embedding.T))
+
+        label = torch.tensor(np.array(ratings)).to(device)
+
+        return pred, label
```
joneswong, line 33: when do you calculate these for negative examples?

joneswong:
@DavdGao hi dawei, please help us check the changes about datasets. Thanks!

federatedscope/mf/dataset/movielens.py
```diff
@@ -75,7 +97,16 @@ def __init__(self, root, num_client, train_portion=0.9, download=True):
                                "You can use download=True to download it")
 
         ratings = self._load_meta()
-        self._split_n_clients_rating(ratings, num_client, 1 - train_portion)
+
+        self.processed_data = os.path.join(self.root, self.base_folder,
+                                           'processed_data.pkl')
+        if os.path.exists(self.processed_data):
```
DavdGao, line 103: I don't think line 104-105 will be executed, since a new exp directory (`self.root`) will be created for each time.

federatedscope/mf/model/model.py
DavdGao, line 18: What's the difference between `torch.nn.Embedding` and `torch.normal`, and is `torch.nn.Embedding` much faster than `torch.normal`?

DavdGao: Me too. In my opinion, the label matrix for MF task is very sparse, but the user/item embedding  and prediction results should be dense.

federatedscope/mf/model/model.py
```diff
@@ -15,47 +14,34 @@ class BasicMFNet(Module):
...
-
-        return mask * pred, label, float(np.prod(pred.size())) / len(ratings)
+        device = self.embed_user.weight.device
+
+        indices = torch.tensor(np.array(indices)).to(device)
+
+        user_embedding = self.embed_user(indices[0])
+        item_embedding = self.embed_item(indices[1])
+
+        pred = torch.diag(torch.matmul(user_embedding, item_embedding.T))
```
DavdGao, line 29: Why `pred` is calculated by `torch.diag` here?

DavdGao (commented):
Please see the inline comments

federatedscope/mf/model/model.py
joneswong: It is OK to me to just fit the observed entries, but it is a very naive baseline approach I guess

joneswong (approved):
approved.

PR #553, opened 2023-03-27
Modify files to supprot distributed mode for XGBoost

qbc2016:
As the title says.

### DIFF
--- a/federatedscope/vertical_fl/xgb_base/worker/train_wrapper.py
+++ b/federatedscope/vertical_fl/xgb_base/worker/train_wrapper.py
@@ -21,7 +21,7 @@ def wrap_client_for_train(client):
 
         if train_flag == 'train_finish':
             # Report evaluation results
-            self.eval_finish_flag = False
+            self.eval_finish_flag = 0
             self.eval(tree_num)
             self._check_eval_finish(tree_num)
         elif train_flag == 'call_for_node_split':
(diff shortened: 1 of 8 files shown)


federatedscope/vertical_fl/xgb_base/worker/train_wrapper.py
```diff
@@ -62,7 +62,7 @@ def _find_and_send_split(self, split_ref, tree_num, node_num):
                 self.model[tree_num][node_num].member = client_id
 
                 split_ref['feature_idx'] -= accum_dim
-                split_child = False
+                split_child = 0
```
xieyxclack, line 65: I suggest preserving `True/False` here to make it readable, and only transforming the type when `split_child` is in `message.content`

scripts/distributed_scripts/distributed_configs/distributed_xgb_client_1.yaml
xieyxclack, line 1: When using distributed mode, we need to support the case where adopted datasets are distributed (load from different files)

xieyxclack (commented):
LGTM, please refer to inline comments for suggestions

xieyxclack (approved):
LGTM. 
TODO: to enable multiple files in distributed mode, which is provided in the previous version but has bugs now. We will fix the bugs ASAP.

PR #554, opened 2023-03-28
Add label-based version for RF

qbc2016:
As the title says, the idea is that task party encrypts indicator and label, decrypts partial sums to compute gain for RF


xieyxclack (approved):
LGTM

PR #555, opened 2023-03-28
Compression methods

xieyxclack:
This pr includes:
- Message compression based on grpc (gzip & deflate)
- Symmetric uniform model compression (32 bit -> 16/8 bit)
- Unittest for model compression in distributed & standalone mode
- Docs of usage of these compression methods

Thanks a lot for the help from @chenhesen!


yxdyc (approved):
LGTM. By the way, it is suggested to add (asymmetric) affine quantization in the next version, as it is a relatively simple and natural extension for the implemented symmetric uniform quant here, while can achieving smaller precision loss.

PR #566, opened 2023-04-03
[HOTFIX] Bug fix for nbafl

DavdGao:
- fix the bug that client in core/client.py doesn't have the attribute "data"; 
- modify unittest for nbafl to inject noise in server


xieyxclack (approved):
LGTM

PR #570, opened 2023-04-10
[DOC] Github actions for Auto API Ref

rayrayraykk:
The API reference can be found [here](https://github.com/alibaba/FederatedScope/suites/12146387326/artifacts/640644695) in https://github.com/alibaba/FederatedScope/actions/runs/4663479590 .


joneswong (approved):
approved.

PR #599, opened 2023-05-16
LLM Enhancement: model sharding

rayrayraykk:
1. Remove warnings in transformers
2. Some features of `accelerate` are experimental.
3. Enable model sharding & offloading when set `cfg.llm.accelerator.use = True` to all available GPUs.

Try new features with:
```python
from federatedscope.core.auxiliaries.data_builder import get_data
from federatedscope.core.auxiliaries.model_builder import get_model
from federatedscope.core.configs.config import global_cfg
from federatedscope.core.auxiliaries.trainer_builder import get_trainer
from federatedscope.core.monitors.monitor import Monitor


def main():
    init_cfg = global_cfg.clone()
    init_cfg.merge_from_file('federatedscope/llm/baseline/testcase.yaml')

    init_cfg.model.type = 'decapoda-research/llama-7b-hf@huggingface_llm'
    # Create a LLM model, 
    # `model.type` = 'decapoda-research/llama-7b-hf@huggingface_llm'
    model = get_model(init_cfg)

    # Create a `ClinetData`
    data, modified_cfg = get_data(config=init_cfg.clone())
    client_data = data[1]
    init_cfg.merge_from_other_cfg(modified_cfg)

    init_cfg.freeze()

    # Create an LLM Trainer instance, `trainer.type` = 'llmtrainer'
    trainer = get_trainer(model=model,
                          data=client_data,
                          device='cuda:0',  # Invalid for model sharding
                          config=init_cfg,
                          monitor=Monitor(init_cfg))

    # Start training
    for i in range(100):
        num_sample, state_dict, res = trainer.train()
        trainer.update(state_dict)
        print(f'Round {i}: {res}')

    # Evaluate
    eval_res = trainer.evaluate()
    print(f'Evaluation: {eval_res}')


if __name__ == "__main__":
    main()
```

### DIFF
--- a/federatedscope/llm/trainer/trainer.py
+++ b/federatedscope/llm/trainer/trainer.py
@@ -66,6 +66,7 @@ class LLMTrainer(GeneralTorchTrainer):
     def _hook_on_batch_backward(self, ctx):
         ctx.optimizer.zero_grad()
         if self.use_accelerator:
+            # TODO: enable `accelerator.accumulate(model)`
             self.accelerator.backward(ctx.loss_task)
         else:
             ctx.loss_task.backward()


qbc2016 (approved):
LGTM.

federatedscope/llm/trainer/trainer.py
```diff
@@ -36,9 +63,21 @@ def _hook_on_batch_forward(self, ctx):
         ctx.loss_batch = CtxVar(loss, LIFECYCLE.BATCH)
         ctx.batch_size = CtxVar(len(labels), LIFECYCLE.BATCH)
 
+    def _hook_on_batch_backward(self, ctx):
+        ctx.optimizer.zero_grad()
+        if self.use_accelerator:
```
yxdyc, line 68: Current version is OK for non-accumulated gradients scenarios. While we should keep in mind that we need properly use `accelerator.accumulate(model)`, which can be a TODO item, see more details [here](https://huggingface.co/docs/accelerate/concept_guides/gradient_synchronization)

federatedscope/llm/trainer/trainer.py
```diff
@@ -36,9 +63,21 @@ def _hook_on_batch_forward(self, ctx):
         ctx.loss_batch = CtxVar(loss, LIFECYCLE.BATCH)
         ctx.batch_size = CtxVar(len(labels), LIFECYCLE.BATCH)
```
yxdyc, line 64: What should be the definition of value of `ctx.batch_size` when we using accelerator? In megetron-LM, there are micro-batch-size and global-batch-size, see similar discussion in [here](https://huggingface.co/docs/accelerate/concept_guides/performance#observed-batch-sizes)

yxdyc (commented):
LGTM, see the minor questions in inline comments

PR #610, opened 2023-05-24
[Feature] Offsite tuning

rayrayraykk:
How to use:
```bash
python federatedscope/main.py --cfg federatedscope/llm/baseline/testcase.yaml llm.offsite_tuning.use True llm.accelerator.use False
```

### DIFF
--- a/federatedscope/llm/offsite_tuning/utils.py
+++ b/federatedscope/llm/offsite_tuning/utils.py
@@ -15,6 +15,10 @@ logger = logging.getLogger(__name__)
 
 
 def get_layers(adapter_model):
+    """
+    Modified from the official implementation:
+    https://github.com/mit-han-lab/offsite-tuning/tree/main
+    """
     if isinstance(adapter_model.model, OPTForCausalLM):
         layers = adapter_model.model.model.decoder.layers
     elif isinstance(adapter_model.model, GPT2LMHeadModel):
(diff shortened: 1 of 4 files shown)


federatedscope/core/configs/cfg_llm.py
```diff
@@ -32,6 +32,13 @@ def extend_llm_cfg(cfg):
     cfg.llm.adapter.use = False
     cfg.llm.adapter.args = []
 
+    # ---------------------------------------------------------------------- #
+    # Offsite-tuning related options
+    # ---------------------------------------------------------------------- #
+    cfg.llm.offsite_tuning = CN()
+    cfg.llm.offsite_tuning.use = False
+    cfg.llm.offsite_tuning.strategy = 'drop_layer'
+
```
qbc2016, line 41: Maybe we should move the above cfgs to configs/cfg_llm.py

federatedscope/llm/offsite_tuning/utils.py
```diff
@@ -0,0 +1,115 @@
...
+    OPTForCausalLM,
+    GPT2LMHeadModel,
+    BloomForCausalLM,
+)
+from federatedscope.llm.model.adapter_builder import AdapterModel
+
+logger = logging.getLogger(__name__)
+
+
+def get_layers(adapter_model):
```
yxdyc, line 17: If these codes are from their official codes, please point out it such as 
`most of the codes are modified from the official implementation`. After we adding more novel implementations, we can remove this.

PR #625, opened 2023-06-07
Fix NaN in LLM train

rayrayraykk:
We track the ignore calculation loss caused by ignore_index=- 100 in CROSSENTROPYLOSS. One reason for this is that the `response` is empty, and another reason (often found in dolly-15k) is that the LLMDataset causes the input_max_len to exceed the max_len of the tokenizer and therefore ignores the `response`.

When `llm.tok_len: 1000`, dolly contains 177 abnormal samples，and alpaca contains 28 abnormal samples.

image


qbc2016 (approved):
LGTM.

PR #626, opened 2023-06-07
Optimize gsm8k evaluation

rayrayraykk:
How to use:
```bash
nohup python federatedscope/llm/eval/eval_for_gsm8k/eval.py --cfg federatedscope/llm/baseline/llama.yaml adapter.use False device 1 >gsm8k_test.log &
```


qbc2016 (approved):
LGTM.

PR #627, opened 2023-06-08
Add code search net for SFT

rayrayraykk:
Due to the size of code search net is huge, we support subsample for it.
How to use:
1. Download from huggingface `python federatedscope/llm/dataset/code_search_net.py`
2. Run `python federatedscope/main.py --cfg federatedscope/llm/baseline/llama.yaml data.type code_search_net@llm data.subsample 0.001`


qbc2016 (approved):
LGTM.

PR #634, opened 2023-06-15
fix bug: _param_filter(backend:pytorch)

variant-star:
1. in _core.trainers.trainer.py#393-395_, the return in the _else_ case should not be a _lambda function_ but a _bool_ value.

2. in _core.trainers.trainer.py#399_, with _pytorch_ as backend, because _self.ctx.trainable_para_names_ store the keys of model's state_dict (refer to _core.auxiliaries.model_builder.get_trainable_para_names#206_), key of parameters should be fed into _trainable_filter_

I'm not familiar with _tensorflow_ but it seems that such modification is not compatible with _tensorflow_ backend.

### DIFF
--- a/federatedscope/core/workers/server.py
+++ b/federatedscope/core/workers/server.py
@@ -950,16 +950,15 @@ class Server(BaseServer):
         if isinstance(content[1], list):  # multiple model
             sample_size = content[0]
             re_completed_model = [
-                copy.deepcopy(self.models[i].state_dict()) for i in len(content[1])
+                copy.deepcopy(self.models[i].state_dict()) for i in range(len(content[1]))
             ]
-            for i in len(content[1]):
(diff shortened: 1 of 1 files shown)


variant-star:
Hi all, as I described in this PR, I'm now pretty sure there is a bug in __param_filter_. When I tried to simply fix the bug, more and more errors were raised and I couldn't pass the unit tests. Can anyone help take a look?

xieyxclack:
> Hi all, as I described in this PR, I'm now pretty sure there is a bug in __param_filter_. When I tried to simply fix the bug, more and more errors were raised and I couldn't pass the unit tests. Can anyone help take a look? @rayrayraykk

Thanks a lot for your contribution, se will see what happens in this pr and help you!

PR #637, opened 2023-06-15
[Hotfix]Save best model

### LINKED ISSUE #636: The saved model is not the best model from validation set
The model is saved in the final round with latest state dict.

rayrayraykk:
Fix #636: The model is saved in the final round with latest state dict.


qbc2016 (approved):
LGTM.

Issue #639, opened 2023-06-16
Trainable parameters are not properly filtered


rayrayraykk:
This is an issue to log bug fixes related to #634.

Known bugs:
1. ~~trainable parameters to be not properly filtered in `trainer.py`~~
2. ~~`get_trainable_para_names` are not real trainable param in `context.py`~~
3. ~~when using global evaluation the track_running_stats won't be updated in the server which will result in bad results~~
4. ~~bad results in `gaussian_noise` attack~~
5. ~~weird loss and bad results in async~~ (caused by model delta: init_model - update, where init_model should filter running_stats.)

PR #642, opened 2023-06-19
fix bug: _param_filter

rayrayraykk:
This PR is to solve the problem in #634.
Track the change log in #639.

**Notice**: due to this change, `trainer.get_model_para()` will **ONLY** return trainable param.

### DIFF
--- a/federatedscope/core/workers/server.py
+++ b/federatedscope/core/workers/server.py
@@ -697,8 +697,6 @@ class Server(BaseServer):
         # We define the evaluation happens at the end of an epoch
         rnd = self.state - 1 if msg_type == 'evaluate' else self.state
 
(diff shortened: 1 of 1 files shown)


yxdyc (dismissed):
LGTM. IMHO, some todos are user-dependent thus we may should choose the comprehensive (but perhaps redundant) ones as the default option, such that user customized changes can be incremental. E.g., sending full model parameters is OK as users can easily add post-operations to filter out the non-trainable ones.

Just as the open-closed principle (OCP) states 
> software entities (classes, modules, functions, etc.) should be open for extension, but closed for modification

yxdyc (approved):

xieyxclack (approved):
LGTM

PR #644, opened 2023-06-19
functionality merge for LLM

sundave1998:
please refer to individual commit comments for details

### DIFF
--- a/federatedscope/core/auxiliaries/logging.py
+++ b/federatedscope/core/auxiliaries/logging.py
@@ -158,8 +158,8 @@ def init_wandb(cfg):
                entity=cfg.wandb.name_user,
                config=cfg_yaml,
                group=dataset_name,
-               job_type=method_name,
-               name=exp_name,
+               job_type=exp_name,
+               name=method_name,
                notes=f"{method_name}, {exp_name}")
 
 
--- a/federatedscope/core/configs/cfg_training.py
+++ b/federatedscope/core/configs/cfg_training.py
@@ -45,6 +45,9 @@ def extend_training_cfg(cfg):
 
     # when model is too large, users can use half-precision model
     cfg.train.is_enable_half = False
+        
+    # check whether model uses accelerator
+    cfg.train.use_accelerator = False
 
     # ---------------------------------------------------------------------- #
     # Finetune related options
--- a/federatedscope/core/configs/cfg_training.py
+++ b/federatedscope/core/configs/cfg_training.py
@@ -45,7 +45,7 @@ def extend_training_cfg(cfg):
 
     # when model is too large, users can use half-precision model
     cfg.train.is_enable_half = False
-        
+
     # check whether model uses accelerator
     cfg.train.use_accelerator = False
 
--- a/federatedscope/core/fed_runner.py
+++ b/federatedscope/core/fed_runner.py
@@ -461,7 +461,8 @@ class StandaloneRunner(BaseRunner):
         server_msg_cache = list()
         while True:
             if len(self.shared_comm_queue) > 0:
-                msg = self.shared_comm_queue.popleft()
+                # msg = self.shared_comm_queue.popleft()
+                msg = self.shared_comm_queue[0]
                 if not self.cfg.vertical.use and msg.receiver == [
                         self.server_id
                 ]:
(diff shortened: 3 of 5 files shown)


federatedscope/llm/model/adapter_builder.py
```diff
@@ -17,13 +17,20 @@ def enable_adapter(model, package, adapter, **kwargs):
...
-            r = kwargs.get('lora_r', 8)
+            r = kwargs.get('r', 8)
             lora_alpha = kwargs.get('lora_alpha', 32)
             lora_dropout = kwargs.get('lora_dropout', 0.1)
+            target_modules = kwargs.get('target_modules', None)
             peft_config = LoraConfig(r=r,
+                                     target_modules=target_modules,
                                      lora_alpha=lora_alpha,
                                      lora_dropout=lora_dropout)
+            print("using peft config:", peft_config)
```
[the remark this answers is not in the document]
sundave1998: updated related file and replaced print with logging, thank you.

yxdyc (commented):
Plz help us to double check the correctness of modifications w.r.t. the queue & heap in `fed_runnner` @xieyxclack.  Others look good to me.

federatedscope/core/auxiliaries/logging.py
```diff
@@ -158,8 +158,8 @@ def init_wandb(cfg):
                entity=cfg.wandb.name_user,
                config=cfg_yaml,
                group=dataset_name,
-               job_type=method_name,
-               name=exp_name,
+               job_type=exp_name,
```
xieyxclack, line 161: Why do we need to exchange `exp_name` and `method_name`?

federatedscope/core/configs/cfg_training.py
```diff
@@ -46,6 +46,9 @@ def extend_training_cfg(cfg):
     # when model is too large, users can use half-precision model
     cfg.train.is_enable_half = False
 
+    # check whether model uses accelerator
+    cfg.train.use_accelerator = False
```
xieyxclack, line 50: what is this args used for?

federatedscope/core/fed_runner.py
```diff
@@ -461,7 +461,8 @@ def _run_simulation(self):
         server_msg_cache = list()
         while True:
             if len(self.shared_comm_queue) > 0:
-                msg = self.shared_comm_queue.popleft()
+                # msg = self.shared_comm_queue.popleft()
```
xieyxclack, line 464: Why do we need to do the `pop` operation after handling the message here? Can you give an example?

xieyxclack (commented):
Please refer to the inline comments for suggestions! Thank you

federatedscope/core/configs/cfg_training.py
sundave1998: I have implemented additional packages to be used in the application of differential privacy in FedScope, where [accelerator](https://github.com/huggingface/accelerate) is useful.

federatedscope/core/fed_runner.py
sundave1998: This modification was applied to increase robustness of the pipeline, in both standalone and distributed settings, it is generally a better practice (in my opinion) to remove the query only after it has been fully handled. 
One example would be for a distributed algorithm where one client process has been killed during training, in this case, the run can continue when the client restarts and receive the message again. 
Another example is when the standalone process is paused (such as in Jupyter), the experiment can continue by rerunning function `runner.run()`, the algorithm will continue from the last message still left in queue.
On the other hand, if the message is first popped then handled, the runner is unable to recover, since it reads an empty queue and considers the run finished.

federatedscope/core/auxiliaries/logging.py
sundave1998: this is only related to the viewing logging in wandb web UI. in my exps (I believe applicable to most LLMs experiments) generally the config yaml is modified but the optimizer (corresponding to method_name) is not. In the wandb API, the job_type Is listed as separate colored experiments where the different names are different runs of the same color group. If the changes seem unnecessary, they can be reverted.

federatedscope/core/fed_runner.py
xieyxclack: - `_run_simulation` is only used in standalone mode.
- Please double check the case in Jupyter @rayrayraykk 
- The main idea behind the original implementation (i.e., pop first and handle later) is to ensure the popped msg is exactly the one you handled, since during the handling process some new message might be generated and pushed into the comm_queue

federatedscope/core/auxiliaries/logging.py
yxdyc: This change will affect some historical experiments (e.g. scripts in our benchmarks). IMHO, it only affects the presentation of the results, but not the correctness of the results. That is, the user will see the grouped results on the WandB page in a way that is inconsistent with some historical result. If the user wants to change it, the cost should be very small.

In general I agree with this change, but a comment could be added to simply remind the user, e.g. "in [PR#xx]](link goes here), the `exp_name` and `method_name` were exchanged for a better grouping manner"

PR #649, opened 2023-06-28
Fix issues in offsite tuning

rayrayraykk:
From @HarliWu :
Some bugs have been fixed for /llm/offsite_tuning/. Specifically, the keys of "emulator and adapter" mismatch that of the raw model.
Also, we optimize the RAM usage of the training of the offsite-tuning.
Support the loading of offsite_tuning model in FSChatBot

Minor changes are made from #647 .


qbc2016 (approved):
LGTM.

PR #650, opened 2023-06-28
Optimize Standalone memory usage in offsite-tuning

rayrayraykk:
When setting `federate.share_local_model` to True in offsite-tuning, we could save a lot of memory!


ZiTao-Li (approved):
LGTM

PR #653, opened 2023-07-03
[Experimental Feature]DeepSpeed for LLM with standalone and distributed

rayrayraykk:
TODO: 
* should clients share the same `ctx.model_engine`? (compatibility of `share_local_model` and `Deepspeed`)


qbc2016 (approved):
LGTM.

PR #675, opened 2023-08-02
Fix and update distillation

rayrayraykk:
The teacher is a complete emulator. The correct implementation is that the student is a compressed emulator, and the incorrect implementation is that student was compressed emulator+adapter.


qbc2016 (approved):
LGTM.

PR #683, opened 2023-08-23
 experimental feature: MeZO optimizer

zhenqincn:
This branch contains the implementation of integrating MeZO optimizer.


zhenqincn:
> Since the train and eval will use the same hook_func, we should add an if-else when eval only needs one forward, right?
> 
> And I suggest avoiding `@staticmethod` when the methods in a class need to call each other. It makes the code less readable and maintainable. You can use regular instance methods or class methods instead.

Thanks for these valuable suggestions. In the latest commit, these suggestions have been adopted.

PR #692, opened 2023-09-05
Add HumanEvalX for eval

rayrayraykk:
There is still a bug in HumanEvalX, this feature can be enabled after merging https://github.com/THUDM/CodeGeeX/pull/214, or you can fix it in your docker images


qbc2016 (approved):
LGTM.

PR #693, opened 2023-09-06
build paper list for fl-llm

qbc2016:
As the title says.


xieyxclack:
Good job! Maybe we can also merge it into master branch?

PR #697, opened 2023-09-06
add retry when loss is NaN in train and finetune

rayrayraykk:
add retry when loss is NaN in train and finetune when `cfg.llm.retry_on_nan_loss = True`


qbc2016 (approved):
LGTM.

